PR #4, opened 2018-07-15
docs: improve README.md

watson:


README.md
```diff
@@ -20,7 +21,7 @@ branch](https://github.com/elastic/apm-nodejs-http-client/tree/opbeat).
 ## Installation
 
 ```
-npm install elastic-apm-http-client
+npm install elastic-apm-http-client --save
```
Qard: Is there a particular reason for adding `--save`? It's the default behavior, so I don't think we need that. 🤔

watson: Old habit I guess. I have disabled automatic saving in my personal .npmrc because it's often in the way (I often install a lot of modules for debugging purposes, and I don't want those added to the package.json). But I'm probably the odd one out. If we decide not to have `--save` here, we should probably also remove it from the agent docs.

Qard: I'm fine with including it. I just didn't really get _why_. I don't disable it, personally. I just use `--no-save` when installing something for debugging.

watson: After going back and forth on this for a bit, I think I'd like to keep it. Even though it's now the default behavior, it signals to the reader that this module is intended to be installed as a dependency - as opposed to as a dev-dependency or globally or temporarily for debugging. I like being explicit 😄

Qard: Sounds good. 👍

PR #5, opened 2018-07-16
POC: Support v2 of the APM Server intake API

watson:
This PR adds support for the upcoming new version of the APM Server intake API.

Note that the PR is towards a new `api-v2` branch as this is still experimental and not supported by the APM Server.

BREAKING CHANGE: The client no longer supports v1 of the intake API and the client API have changed as well.

## Todo

- [x] Address all todo's in code
- [x] Update stream-chopper dependency to point to npm when stream-chopper is published


index.js
```diff
@@ -1,75 +1,154 @@
...
-    serverTimeout: opts.serverTimeout
+    pump(stream, zlib.createGzip(), req, function (err) {
+      if (err) onerror(err)
+      // Add listener for req errors in case a timeout occurs AFTER the request
+      // finishes
+      //
+      // TODO: This will produce two errors, which might not be that nice
+      // - First one is caught by the if-sentence above (premature closure)
+      // - Second one is caught by the listener below (socket hung up)
+      req.on('error', onerror)
```
Qard: Can you explain why the error event handler is attached here rather than before the pump?

test/test.js
```diff
@@ -0,0 +1,446 @@
...
+    req.on('data', function () {
+      stream.end()
+    })
+
+    req = processReq(req)
+    req.on('data', function (obj) {
+      t.deepEqual(obj, {req: 2}, 'should get data')
+    })
+    req.on('end', function () {
+      t.ok(true, 'should end request')
```
Qard: `t.pass(msg)` is probably better here.

test/test.js
```diff
@@ -0,0 +1,446 @@
...
+  const server = APMServer(function (req, res) {
+    req.resume()
+    req.on('end', function () {
+      res.end()
+    })
+  }).client({serverTimeout: 1000}, function (stream) {
+    start = Date.now()
+    stream.write({foo: 42})
+  }, function (err) {
+    count++
```
Qard: You could do this cleaner with:

test/test.js
```diff
@@ -0,0 +1,446 @@
...
+    t.equal(err.code, 'DEPTH_ZERO_SELF_SIGNED_CERT')
+    server.close()
+    t.end()
+  })
+})
+
+test('allow unauthorized TLS if asked', function (t) {
+  t.plan(1)
+  const server = APMServer({secure: true}, function (req, res) {
+    t.ok(true, 'should let request through')
```
Qard: `t.pass(msg)` is probably better here.

index.js
watson: The `pump` module will also catch the first error. So if we add it before the pump we'll get the first error twice. But I agree this isn't nice. But that's more down to the fact that the core http module is a bit of a mess at times. It can emit more than one error on the `req` object - even after the request have finished. The 2nd error it emits is actually not coming from the request, but it just uses that object to emit the error on 🤷‍♂️

test/test.js
watson: Ah yes, thanks

test/test.js
watson: fixed 😃

test/test.js
watson: fixed 😃

test/test.js
watson: fixed 😃

index.js
watson: If you have a better idea I'm all 👂s

watson:
@Qard I'd like your input on the TODO's when you have time 😃

index.js
Qard: Perhaps rather than running `onerror` from pump it could _only_ do that via a `req.on('error, onerror)` outside the pump call? Then the pump callback would _only_ call `next`. That should result in only one error path, right? Does pump forward the actual error events across the streams, or does it only aggregate the errors into the callback?

index.js
```diff
@@ -1,75 +1,162 @@
...
+  const serialize = ndjson.serialize()
+  // TODO: Is there a race condition between the `time` config option and the request timeout option?
+  const chopper = new StreamChopper(opts).on('stream', onStream(opts, onerror))
+
+  // All sockets on the agent are unreffed when they are created. This means
+  // that when those are the only handles left, the `beforeExit` event will be
+  // emitted. By listening for this we can make sure to end the requests
+  // properly before exiting. This way we don't keep the process running until
+  // the `time` timeout happens.
+  // TODO: If you start too many clients (e.g. in a test), this will emit a too-many-listeners warning
```
Qard: You could stuff the clients in an array and only attach this handle once at the global scope, iterating over the clients and running a destructor of sorts to do this cleanup.

index.js
```diff
@@ -1,75 +1,162 @@
...
+  // process
+  normalized.serverUrl = parseUrl(normalized.serverUrl)
+  normalized.transport = require(normalized.serverUrl.protocol.slice(0, -1)) // 'http:' => 'http'
+  normalized.agent = getAgent(normalized)
+
+  return normalized
+}
+
+function getAgent (opts) {
+  // TODO: Consider use of maxSockets and maxFreeSockets
```
Qard: Maybe just pass those through?

index.js

index.js
watson: Much better idea. Fixed 👍

index.js
```diff
@@ -1,75 +1,162 @@
...
+      // Sockets will automatically be unreffed by the HTTP agent when they are
+      // not in use by an HTTP request, but as we're keeping the HTTP request
+      // open, we need to unref the socket manually
+      socket.unref()
+    })
 
-  var url = parseUrl(opts.serverUrl || 'http://localhost:8200')
+    // TODO: It doesn't really make sense to set req.serverTimeout on a
+    // streaming request before all the data is sent. We need find a solution
+    // for how to handle timeout of the response.
```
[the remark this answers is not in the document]
watson: I was thinking about that, but it leaves open a blind spot from when the request have been sent to the response starts coming in. For now I've just added a warning in the readme not to set the `serverTimeout` below the `time` timeout.

index.js
watson: Great idea. Fixed 😃

watson:
@Qard pushed lots of updates:

- Addressed all your review comments
- Client is now an EventEmitter (can emit `error` and `finish` - no need to pass in separate `onerror` callback any more)
- Calling `Client()` no longer returns a writable stream (instead a client instance is returned)
- Wrapped logic for sending spans, transactions, and errors into 3 functions: `sendSpan()`, `sendTransaction()`, and `sendError()`. These automatically wrap the given object in the transport envelope
- Added `client.flush()` function to flush the internal buffer
- Added `client.end()` function to end the client gracefully
- Added `client.destroy()` function to destroy the client an all underlying resoures

Things I'm still considering:

- The few TODO's that are left
- Moving truncating logic (e.g. span name can not be longer than 1024 bytes) from the agent to this module

Please let me know what you think now 😃

Qard:
Any reason not to just make `Client` an actual stream? There's a bunch of functions on it duplicating the interface of a stream, but it's not a "real" stream. 🤔

watson:
The reason why I dropped treating it as a writable stream is that you're not supposed to write to it anymore. Instead of calling `write` with the data, you interact with it using 3 higher level functions: `sendSpan`, `sendTransaction`, and `sendError`. So it's not really behaving like a stream - it's more like a high level wrapper around a stream

But maybe there's a way to turn it into a writable stream in a clean way that I hadn't thought about?

watson:
Just fixed the last big TODO item. The two remaining are simple if we decide upon the right action

Qard:
It seems reasonable to me for it to remain a stream and for the `send*` methods to just be convenience methods to wrap the objects with their appropriate keys. That's _basically_ all they are currently, but with the minor disadvantage that we now can't treat the object like a _real_ stream.

watson:
@Qard I changed it to a writable stream now. It actually made a lot of sense. Thanks 😃

watson:
Also fixed the last TODO's

README.md
```diff
@@ -99,9 +99,19 @@ Streaming configuration:
...
 Error argument when called.
 
+The client cannot recover from errors.
+
+### Event: `warning`
+
+Emitted if an error occurs. The listener callback is passed a single
+Error argument when called.
+
+The client can receover from warnings.
```
Qard: s/receover/recover/

index.js
```diff
@@ -70,63 +72,69 @@ Client.prototype._ref = function () {
...
-  if (this._ended) return // TODO: call bacllback?
 
-  if (cb && this._active) {
-    // if there's an outgoing http request, queue the callback to be called
-    // when we're sure that the data has left the system
-    this._onflushed = cb
-    this._chopper.chop()
-  } else {
-    this._chopper.chop(cb)
-  }
```
Qard: What's the reason for moving this into the `_write` logic?

index.js
```diff
@@ -70,63 +72,69 @@ Client.prototype._ref = function () {
...
   clients[this._index] = null // remove global reference to ease garbage collection
-  if (cb) this.once('finish', cb)
   this._ref()
   this._stream.end()
+  cb()
 }
 
-Client.prototype.destroy = function () {
+// Overwrite destroy instead of using _destroy because readable-stream@2 can't
+// be trusted
```
Qard: Can you provide some more context here?

Qard:
Other than some minor nits, LGTM.

index.js
watson: We write the special "flush" signal so that the order of writes and flushes are kept. If we where to just flush the client right here, the internal Writable buffer might still contain data that hasn't yet been given to the _write function.

I've added a comment to explain this.

index.js
watson: After a stream is destroyed, we want a call to either `client.write()` or `client.end()` to both emit an error and call the provided callback. Unfortunately, in readable-stream@2 and Node.js <10, this is not consistent. This has been fixed in Node.js 10 and will also be fixed in readable-stream@3.

I've updated the comment to better explain this.

watson: Specifically, this is because calling `stream.destroy()` emits `finish` synchronously in Node.js <10 and in readable-stream@2. So when calling `stream.end(callback)` later, the `callback` is attached to the `finish` event, but since it has already fired, nothing happens.

In Node.js 10 and in the upcoming readable-stream@3, this have been fixed so that the `finish` event isn't emitted until you call `stream.end()`.

You can see an isolated test case with the problem here: https://gist.github.com/watson/6afbb67423c979a5c8f311f137184320

README.md
watson: Fixed

watson:
@Qard I fixed all your review comments.

Added single TODO where I'm not sure if there's a difference between checking `_writableState.ended` or `_writableState.ending` for if the stream has ended. I chose `ending` as that sounds more right. But I need to double check.

I also dropped the concept of the `warning` event as it was hard to say if an `error` was recoverable just by looking at the error event. Instead the client should now emit `close` in case it couldn't recover. So that's what you have to listen for to see see if there's a problem. Now all errors are emitted as the `error` event.

watson:
What I still want to do:

- Move metadata generation into this module
- Move truncation into this module
- Add tests for and document the `drain` event

But I think I'll do that in other PR's once this is merged (in fact I already have a [branch](https://github.com/watson/apm-nodejs-http-client/tree/v2-1) where I'm working on the first item on the list).

PR #10, opened 2018-08-08
perf: use faster ndjson algorithm

watson:


watson:
Yeah, it's just super convenient for testing. Having to implement the parsing algorithm our selfs is not worth it in my opinion.

PR #13, opened 2018-08-10
feat: add write buffer to improve performance

watson:


Qard:
Unfortunately the encoding symbol approach means the `write` method is no longer directly usable. Not sure how much we should consider that. 🤔

index.js
```diff
@@ -16,10 +16,11 @@ const pkg = require('./package')
...
 const flush = Symbol('flush')
-const metadataSym = Symbol('metadata')
-const transactionSym = Symbol('transaction')
-const spanSym = Symbol('span')
-const errorSym = Symbol('error')
+
+Client.metadataEnc = Symbol('metadata')
+Client.transactionEnc = Symbol('transaction')
+Client.spanEnc = Symbol('span')
+Client.errorEnc = Symbol('error')
```
Qard: What about something like this?

```js
const encoding = {
  metadata: Symbol('metadata'),
  transaction: Symbol('transaction'),
  span: Symbol('span'),
  error: Symbol('error')
}
Client.encoding = encoding
```

Less repetitive and more clear, in my opinion, than `*Enc`.

Qard: I'd probably also `Object.freeze` the object and upper-case names to indicate constants.

test/writev.js
```diff
@@ -0,0 +1,108 @@
...
+        t.equal(err.message, 'write called on destroyed Elastic APM client')
+        server.close()
+        t.end()
+      })
+
+      client[sendFn]({req: 1})
+      client[sendFn]({req: 2})
+
+      // Destroy the client before the _writev function have a chance to be called
+      client.destroy()
```
watson: @Qard I'm actually not sure this is a feature, but instead a bug? What would you expect to happen in this case? Should just abort writing without emitting the error?

Qard: I'm not sure what you are trying to do here...

Qard: It's difficult to reason about the exact timing...

Is it possible that it might get aborted part way through a `_writev`? As in, some of the data has been written, but not all? How would we want to react to an abort? Retrying? Dropping the data?

I don't really have an informed opinion on what to do about this right now.

watson: I thought about it some more and decided to swallow this error

index.js
```diff
@@ -105,7 +105,9 @@ Client.prototype._write = function (obj, enc, cb) {
 
 Client.prototype._writev = function (objs, cb) {
   if (this._destroyed) {
-    this.emit('error', new Error('write called on destroyed Elastic APM client'))
+    // We should only get to point if the client was destroyed in the same tick
```
Qard: to this point*

index.js
```diff
@@ -105,7 +105,9 @@ Client.prototype._write = function (obj, enc, cb) {
...
+    // as data was written to a corked client. To now trip users up about this,
```
Qard: s/now/not/

index.js
```diff
@@ -105,7 +105,9 @@ Client.prototype._write = function (obj, enc, cb) {
...
+    // let's just silently ignnore it.
```
Qard: ignore*

index.js

index.js

watson:
@Qard wow 😂 I think that's the record of the most typos in one sentence 😄

PR #14, opened 2018-08-31
chore: bump fast-safe-stringify to version ^2.0.6

watson:


Qard:
Tests seem to fail due to different assertion counts.

watson:
@Qard re-ran the tests and they passed. Not sure what was going on

PR #17, opened 2018-09-04
feat: add incremental back-off in case of APM Server issues

watson:


Qard:
It's a bit complicated, so it might be worth re-evaluating at some point. But it looks reasonable for now. 👍

watson:
It's still in progress (I have some weird bugs in the tests that I'm trying to work out). I don't like how complicated it is either, but I put it up here primarily so the other agent devs could see how I did it. I'm not sure it's worth adding at this time though, so for now I'm concentrating on other things.

PR #20, opened 2018-09-13
test: check that invalid serverUrl throws

watson:


watson:
@Qard This PR was technically marked as in-progress. But I guess there's no harm done in merging it if you agree with the behavior

Qard:
Oh, oops. Missed that. 😅 

But yeah, I think having a test to prove that behaviour is good.

PR #25, opened 2018-10-19
refactor: use new chopper transform feature to gzip

watson:
This fixes a bug where the max size of the HTTP body (750kb) was wrongly counted towards the uncompressed bytes. The new version of stream-chopper will count towards the compressed bytes instead.


watson:
Hmm need to just fix these errors

watson:
@Qard this should be ready now

PR #29, opened 2018-11-08
refactor: upgrade to readable-stream@3

watson:


watson:
@mafintosh could I get your 👀 on this one as well 😃

watson:
The reason why the tests fail are because the following assumptions have changed in the tests:

- In rs3, `stream.end()` no longer results in an error being emitted for a destroyed stream?
- In rs3, calling `stream.destroy()` synchronously after calling `stream.write()` will result in an error being emitted because the stream becomes destroyed before the write have a chance to complete

But maybe this is expected behavior?

watson:
@Qard just updated this with a few bug-fixes and now the tests also pass. Can you do a 2nd review?

PR #34, opened 2018-11-26
feat: add Kubernetes data to metadata

watson:
Implementing elastic/apm#21


watson:
🤔 Instead of environment variables, should this be config options that the agent it self then sets based on the environment variables?

index.js
```diff
@@ -459,6 +459,27 @@ function getMetadata (opts) {
...
 
+  const k8sNodeName = process.env.KUBERNETES_NODE_NAME
+  const k8sNamespace = process.env.KUBERNETES_NAMESPACE
+  const k8sPodName = process.env.KUBERNETES_POD_NAME
+  const k8sPodUID = process.env.KUBERNETES_POD_UID
+
+  if (k8sNodeName || k8sNamespace || k8sPodName || k8sPodUID) {
+    payload.kubernetes = {
+      namespace: k8sNamespace,
+      name: undefined,
```
Qard: Should this be `node`?

watson:
@Qard It turns out it might be possible to also get the container id and pod UID from `/proc/self/cgroup` (see https://github.com/elastic/apm-agent-go/pull/342). So by allowing this to be set by config options we make it 100% up to the implementer where the information comes from

index.js

Qard:
Config sounds good to me then. 👍 

I'm starting to wonder though if we should just have a `custom` config option for custom metadata, much like we have `custom` for transaction data. 🤔

watson:
@Qard yeah, might not be a bad idea. For this PR however, I've just updated it so that it uses 4 unique config options

watson:
@Qard do you mean in how we name the config options? What do you propose? I was for a moment also thinking to nest the 4 options under a parent `kubernetes` key

Qard:
Yeah, my thinking was to make the config names match the context object name. Some might not recognize the k8s name.

watson:
@Qard fixed

PR #36, opened 2018-12-04
feat: ensure strings always truncated

Qard:
If not otherwise truncated, all strings should be chopped at 1024 characters.

Fixes: https://github.com/elastic/apm-agent-nodejs/issues/630

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,9 +128,10 @@ Data sanitizing configuration:
   disable truncation. This applies to the following properties:
   `error.exception.message` and `error.log.message` (default: `2048`
   bytes)
-- `truncateStringsAt` - The maximum size in bytes for souce code
-  lines in stack traces. Lines above this length will be truncated
-  (default: `1000` bytes)
+- `truncateStringsAt` - The maximum size in bytes for strings.
+  Lines above this length will be truncated (default: `1000` bytes)
+- `truncateQueriesAt` - The maximum size in bytes for queries.
+  Lines above this length will be truncated (default: `1000` bytes)
 
 Debug options:
 
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ Data sanitizing configuration:
   bytes)
 - `truncateStringsAt` - The maximum size in bytes for strings.
   String values above this length will be truncated (default: `1024` bytes)
-- `truncateQueriesAt` - The maximum size in bytes for queries.
+- `truncateQueriesAt` - The maximum size in bytes for database queries.
   Lines above this length will be truncated (default: `1000` bytes)
 
 Debug options:
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ Data sanitizing configuration:
 - `truncateStringsAt` - The maximum size in bytes for strings.
   String values above this length will be truncated (default: `1024` bytes)
 - `truncateQueriesAt` - The maximum size in bytes for database queries.
-  Lines above this length will be truncated (default: `1000` bytes)
+  Queries above this length will be truncated (default: `10000` bytes)
 
 Debug options:
 
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ Data sanitizing configuration:
   `error.exception.message` and `error.log.message` (default: `2048`
   bytes)
 - `truncateStringsAt` - The maximum size in bytes for strings.
-  Lines above this length will be truncated (default: `1000` bytes)
+  String values above this length will be truncated (default: `1024` bytes)
 - `truncateQueriesAt` - The maximum size in bytes for queries.
   Lines above this length will be truncated (default: `1000` bytes)
 
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Data sanitizing configuration:
   disable truncation. This applies to the following properties:
   `error.exception.message` and `error.log.message` (default: `2048`
   bytes)
-- `truncateSourceLinesAt` - The maximum size in bytes for souce code
+- `truncateStringsAt` - The maximum size in bytes for souce code
   lines in stack traces. Lines above this length will be truncated
   (default: `1000` bytes)
 
--- a/lib/truncate.js
+++ b/lib/truncate.js
@@ -1,5 +1,6 @@
 'use strict'
 
+var breadthFilter = require('breadth-filter')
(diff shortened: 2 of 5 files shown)


watson:
Better rebase with master to run this PR against the new circular dependency tests

watson:
As mentioned on Slack, lets include the following in this PR:

Remove the `truncateSourceLinesAt` config option (makes this PR a breaking change) and instead introduce a `truncateStringsAt` config option, so this limit can be controlled.

Additionally `span.context.db.statement` should not be truncated based on `truncateStringsAt`, but instead based on a new config option we'll create called `truncateDatabaseStatementsAt` which should default to `10000`.

README.md
```diff
@@ -128,7 +128,7 @@ Data sanitizing configuration:
   disable truncation. This applies to the following properties:
   `error.exception.message` and `error.log.message` (default: `2048`
   bytes)
-- `truncateSourceLinesAt` - The maximum size in bytes for souce code
+- `truncateStringsAt` - The maximum size in bytes for souce code
```
watson: The description should probably be updated as well 😉

lib/truncate.js
```diff
@@ -8,123 +9,141 @@ exports.span = truncSpan
...
 
 function truncMetadata (metadata, opts) {
-  metadata.process.title = truncate(String(metadata.process.title), opts.truncateKeywordsAt)
+  breadthFilter(metadata, (value, key, path) => {
+    if (typeof value !== 'string') {
+      return value
+    }
+
+    let max = opts.truncateStringsAt
+    if (path.join('.') === 'process.title') {
```
watson: Would it be less expensive to look at `path[0]` and `path[1]` individually?

lib/truncate.js
```diff
@@ -8,123 +9,141 @@ exports.span = truncSpan
 exports.error = truncError
 
 function truncMetadata (metadata, opts) {
-  metadata.process.title = truncate(String(metadata.process.title), opts.truncateKeywordsAt)
+  breadthFilter(metadata, (value, key, path) => {
+    if (typeof value !== 'string') {
+      return value
+    }
+
+    let max = opts.truncateStringsAt
```
watson: Could be implemented with a ternary operator instead. But maybe V8 is able to optimize it to that automatically

lib/truncate.js
```diff
@@ -8,123 +9,141 @@ exports.span = truncSpan
...
+  breadthFilter(trans, (value, key, path) => {
+    if (typeof value !== 'string') {
+      return value
+    }
+
+    let max = opts.truncateStringsAt
+    if (trans.sampled && path[0] === 'context') {
+      max = contextLength(path.slice(1), opts)
+    }
+    switch (path.join('.')) {
```
watson: I think this could just be replaced by `path[0]` right?

README.md
watson: Add docs to `truncateQueriesAt` as well

Qard: fixed

lib/truncate.js
Qard: fixed

lib/truncate.js
Qard: fixed

README.md
Qard: fixed

lib/truncate.js
Qard: fixed

Qard:
The `name`, `type` and `result` fields use the `String(...)` conversion to turn `undefined` into `'undefined'`. I assume that is what those were for. For the other fields, they had an existence check, so those would only have run if the type was a "truthy" value, which I assume was intended to mean it should only run when it is _valid_. Technically in both old an new code, any of those properties could be something falsy and not get converted to a string. 🤷‍♂️

lib/truncate.js
```diff
@@ -8,123 +9,143 @@ exports.span = truncSpan
...
+
+    let max = opts.truncateStringsAt
+    switch (path[0]) {
+      case 'name':
+      case 'type':
+      case 'result':
+        return value
+      case 'context':
+        if (trans.sampled) {
+          max = contextLength(path.slice(1), opts)
```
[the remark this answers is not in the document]
Qard: fixed

lib/truncate.js
```diff
@@ -8,123 +9,143 @@ exports.span = truncSpan
...
+    }
+
+    let max = opts.truncateStringsAt
+    switch (path[0]) {
+      case 'name':
+      case 'type':
+      case 'result':
+        return value
+      case 'context':
+        if (trans.sampled) {
```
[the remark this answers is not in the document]
Qard: fixed

lib/truncate.js
```diff
@@ -8,123 +9,143 @@ exports.span = truncSpan
...
+
+      case 'log':
+        switch (path[1]) {
+          case 'level':
+          case 'logger_name':
+          case 'param_message':
+            max = opts.truncateKeywordsAt
+            break
+          case 'message':
+            if (opts.truncateErrorMessagesAt >= 0) {
```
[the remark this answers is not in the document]
watson: Besides keywords, which _has_ to be truncated at 1024 chars. The main reason to my knowledge we do this is sane defaults and reduced storage. The APM Server will not validate the string lengths on anything besides keywords, so a larger string will go through and will get stored.

@roncohen what do you think? Should we keep allowing the user to set a custom truncation length for error messages (including turning off truncation for error messages completely)?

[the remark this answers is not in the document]
roncohen: sure, that makes sense to me if it's not too complicated and as long as there's sane default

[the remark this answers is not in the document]
Qard: fixed

lib/truncate.js
```diff
@@ -8,123 +9,143 @@ exports.span = truncSpan
...
+
+      case 'exception':
+        switch (path[1]) {
+          case 'type':
+          case 'code':
+          case 'module':
+            max = opts.truncateKeywordsAt
+            break
+          case 'message':
+            if (opts.truncateErrorMessagesAt >= 0) {
```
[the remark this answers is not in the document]
Qard: fixed

PR #42, opened 2018-12-21
feat(options): get cgroups container info

Qard:
Read container info from `/proc/self/cgroup` to get docker container id and kubernetes pod id default values.

### DIFF
--- a/index.js
+++ b/index.js
@@ -415,6 +415,9 @@ function normalizeOptions (opts) {
     if (!normalized.kubernetesPodUID && containerInfo.podId) {
       normalized.kubernetesPodUID = containerInfo.podId
     }
+    if (!normalized.kubernetesPodName && containerInfo.podId) {
+      normalized.kubernetesPodName = hostname
+    }
   }
 
   return normalized
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ const os = require('os')
 const parseUrl = require('url').parse
 const zlib = require('zlib')
 const Writable = require('readable-stream').Writable
+const getContainerInfo = require('container-info')
 const pump = require('pump')
 const eos = require('end-of-stream')
 const streamToBuffer = require('fast-stream-to-buffer')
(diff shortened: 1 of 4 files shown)


watson:
With #34, it was my intention that the 4 Kubernetes config option should be set automatically by the agent and not handled inside this repo. I'm open to moving this responsibility to this repo instead, but I'm not sure if that was what you intented?

Also, out of the 4 config options added in #34, this PR only sets `kubernetesPodUID`. Is that because that's the only variable that could be read from `/proc/self/cgroup`?

watson:
> As for why I put it here, I figured by putting it here one could use the http-client without the agent and still have access to the container data.

Yeah, I guess it can be viewed from two different angles. I was thinking that this lib was low level and wouldn't try to "help" the user by filling these things out. But I can totally see your point. So I'm fine with doing it here.

index.js
```diff
@@ -405,6 +408,15 @@ function normalizeOptions (opts) {
   // process
   normalized.serverUrl = parseUrl(normalized.serverUrl)
 
+  if (containerInfo) {
+    if (!normalized.containerId && containerInfo.containerId) {
+      normalized.containerId = containerInfo.containerId
+    }
+    if (!normalized.kubernetesPodUID && containerInfo.podId) {
+      normalized.kubernetesPodUID = containerInfo.podId
+    }
```
watson: Agreed on call: Add if block checking `!normalized.kubernetesPodName && containerInfo.podId` and if so set the pod name to the hostname

watson:
It seems the tests needs to be updated with the new pod name.

And it would probably be a good idea if we updated the README where the kubernete config options are documented and describe that some of them will get populated automatically based on `/proc/self/cgroup`

watson:
@Qard Did you see this comment?

> And it would probably be a good idea if we updated the README where the kubernete config options are documented and describe that some of them will get populated automatically based on `/proc/self/cgroup`

Qard:
Yep, I'll get to that today. 👍

watson:
Update: We'll just use the os hostname

PR #45, opened 2019-03-02
fix: stringify header values

watson:
Normally headers will always be strings or array of strings, but in case they have been parsed upstream, we must make sure to stringify them here before sending them to the server.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -1,0 +1,27 @@
+'use strict'
+
+exports.context = function (context) {
+  if (!context) return
+
+  if (context.request) stringifyHeaders(context.request.headers)
+  if (context.response) stringifyHeaders(context.response.headers)
+}
+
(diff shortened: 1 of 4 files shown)


lib/stringify.js
```diff
@@ -0,0 +1,27 @@
...
+function stringifyHeaders (headers) {
+  if (!headers) return
+
+  for (const key in headers) {
+    const value = headers[key]
+    if (typeof value === 'string') continue
+    if (Array.isArray(value)) {
+      for (let i = 0; i < value.length; i++) {
+        if (typeof value[i] !== 'string') {
+          value.splice(i, 1, String(value[i]))
```
Qard: Is there a particular reason you use `splice(i, 1, ...)` rather than just `value[i] = ...`?

watson: Let's just pretend you didn't see that 😂 I don't know what I was thinking when I wrote that code

Issue #46, opened 2019-03-07
Truncate Mutates Context Data


blakeembrey:
When using Elastic APM, any data being logged is mutated during the filter execution of `truncate`. This occurred when I tried to use APM with AWS Lambda. The `result` of the lambda function is attached to the context in https://github.com/elastic/apm-agent-nodejs/blob/abb4f1af4a39e59415bf8977608d928e29e11c72/lib/lambda.js#L45, and this client mutates the output object as a result trimming all headers and the body to 1024 bytes. This results in an invalid AWS lambda response.

LucGranato:
This gave me a brain aneurysm.
Lucky that APM was the only thing that was different between the local test and the production environment. It was easier to guess...

PR #49, opened 2019-03-16
Avoid mutation of encoded `obj`

blakeembrey:
Closes https://github.com/elastic/apm-nodejs-http-client/issues/46.

Failing because of the issues mentioned in https://github.com/elastic/apm-nodejs-http-client/issues/46#issuecomment-473490276. Basically, to handle `context.{request,response}.headers` and/or recursion the options are:

1. Stringify the context beforehand and create a new `obj`, failing to properly handle recursion of the top-level object
2. Refactor `breadth-filter` to always run `fn`, then transform the payload of `context.{request,response}.headers` in the callback function
3. Add validation to `context.{request,response}` to ensure the objects match the expected shape - I don't know if anyone is expecting to send non-string headers here


blakeembrey:
Also found a bug in https://github.com/Qard/breadth-filter/issues/7. Fixing it now.

Qard:
I'm working on an update for breadth-filter to support triggering on objects and arrays too. That should enable updating this to work as intended. https://github.com/Qard/breadth-filter/pull/9

Also, I think we would prefer to keep the headers stringification stuff separate from the truncation. Would you mind restoring that bit too?

blakeembrey:
@Qard can you describe what you want? By keeping them separate you run into issue 1, creating new objects to avoid mutation which results in the circular references failing since it’s now a new object.

Qard:
I'll let you know what needs to be done after https://github.com/Qard/breadth-filter/pull/9 lands. The API changes a bit, which enables stuff like replacing circular references with the `[Circular]` value our other tests are expecting, along with enabling direct control of the objects and arrays used to rebuild each given layer in the tree, enabling more fine-grained control over mutability.

As for your question: I just mean the header stringification stuff specifically. The rest of the context should be filtered with the breadthFilter, but the headers object has a known valid structure that we can apply some quicker and easier filters on beforehand. We also want to retain the behavior of making objects within the headers object convert to `[Object object]` as the event will otherwise be rejected by apm server for having an invalid structure.

blakeembrey:
@Qard Got it, I'll be able to split that out once `onObject` exists and check if it's the headers path to apply that logic again. 

Just for my sake, what's the use-case of accepting bad header shapes to start with instead of throwing an error on the JS side?

watson:
@blakeembrey in general the Elastic APM agent should be invisible to the execution of the program. It should never get in the way and have as little of an overhead as possible. Therefore it shouldn't throw any errors either.

Consider this example:

```js
http.createServer(function (req, res) {
  // converts the `session` header value into a session object
  req.headers.session = processSession(req.headers.session)
  if (req.headers.session.loggedIn) {
    // ...
  }
  res.end()
})
```

In this example, the header object is manipulated by the user inside the program. This is perfectly fine, but obviously we don't want an error to be thrown or the APM Server to discard these transactions just because of that. So we need to take into account that someone might have modified the header object. The way we do that is by ensuring that all headers are either strings or arrays of strings.

Qard:
breadth-filter v2 is released now. If you're feeling adventurous, you can update this PR to use the new API. Otherwise, I can make the changes myself. I'm happy to answer any questions you might have about how to use it. It's a bit of a complicated module. 😅

PR #50, opened 2019-03-25
fix(truncate): avoid mutation of encoded `obj`

### LINKED ISSUE #46: Truncate Mutates Context Data

Qard:
@blakeembrey I made an update of your PR to use the latest version of breadth-filter and deal with expected truncation behavior.

Fixes #46
Fixes #49


lib/truncate.js
```diff
@@ -10,192 +10,222 @@ exports.error = truncError
...
-      case 'trace_id':
-      case 'parent_id':
-        max = opts.truncateKeywordsAt
-        break
+      let max = opts.truncateStringsAt
+      switch (path[0]) {
+        case 'name':
+        case 'type':
+        case 'result':
+          value = String(value)
```
watson: No need to stringify the value. It will happen inside `truncate()` as well. So these 3 cases can just be combined with the following 3 cases

lib/truncate.js
```diff
@@ -10,192 +10,222 @@ exports.error = truncError
...
 
-  breadthFilter(span, (value, key, path) => {
-    if (typeof value !== 'string') {
-      return value
-    }
+      let max = opts.truncateStringsAt
+      switch (path[0]) {
+        case 'name':
+        case 'type':
+          value = String(value)
```
watson: No need to stringify the value. It will happen inside `truncate()` as well. So these 3 cases can just be combined with the following group of cases.

lib/truncate.js
```diff
@@ -10,192 +10,222 @@ exports.error = truncError
...
+      }
+
+      return truncate(value, max)
     }
+  })
 
-    return truncate(value, max)
-  }, true)
+  return Object.assign({
+    name: 'undefined',
```
watson: What's the thought behind these `"undefined"` string values?

Qard: This is reproducing the behaviour from before where `trans.name = truncate(String(trans.name), opts.truncateKeywordsAt)` would produce `"undefined"` if the name property was undefined because of the explicit string conversion. The `breadthFilter` will skip the field entirely, if not present as it can't operate on a field it's not aware of.

lib/truncate.js
Qard: fixed

lib/truncate.js
Qard: fixed

Qard:
Made it handle circular arrays and more closely replicate the original behaviours between the different types. I'll see about restoring the split out header string conversion stuff next...

Qard:
Looking closer, it seems splitting the header string conversion stuff back out will be more complicated than I though, because that would just reintroduce the mutation issue. If I put it _before_ the filter, it mutates in-place, if I put it _after_ the filter, a bunch of stuff could miss truncation.

lib/truncate.js
watson: Hm true. I'm wondering about the overhead though. Hopefully these properties will always be present. But lets go with it for now

lib/truncate.js
```diff
@@ -243,3 +267,20 @@ function contextLength (path, opts) {
 
   return opts.truncateStringsAt
 }
+
+function isHeader (path) {
+  return path[0] === 'context' && (path[1] === 'request' || path[1] === 'response') && path[2] === 'headers' && path[3]
+}
+
+function onObjectWithHeaders (value, key, path, isNew) {
+  if (isHeader(path)) return value.toString()
```
watson: Why not use `String(value)` like you do in `truncTransaction` and `truncError`?

Qard: Ah, yeah...good point. Just a second. 😅

Qard: fixed.

PR #53, opened 2019-03-27
feat: allow config options to be updated

watson:
This is a prerequisite for fixing elastic/apm-agent-nodejs#688

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,19 @@ An integer indicating the number of events (spans, transactions, or errors)
 sent by the client. An event is considered sent when the HTTP request
 used to transmit it have ended.
 
+### `client.config(options)`
+
+Update the configuration given to the `Client` constructor. All
+configuration options can be updated except:
+
+- The protocol part of the `serverUrl` (`http` vs `https`)
+- `size`
+- `time`
+- `keepAlive`
+- `keepAliveMsecs`
+- `maxSockets`
+- `maxFreeSockets`
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -191,6 +191,19 @@ An integer indicating the number of events (spans, transactions, or errors)
 sent by the client. An event is considered sent when the HTTP request
 used to transmit it have ended.
 
+### `client.config(options)`
+
+Update the configuration given to the `Client` constructor. All
+configuration options can be updated except:
+
+- The protocol part of the `serverUrl` (`http` vs `https`)
```
Qard: Can we make protocol work too? Without that, it's basically impossible to late-configure it to use https, because the default http serverUrl will get used initially, locking that down. 🤔

watson: I looked into that when I made the PR, but it's much more complicated. So I ended up concluding that it would be simpler to land this PR as is, which would allow for late setting of anything inside metadata (including framework). Then we can make a separate PR to allow delayed start, which would solve this.

Qard:
LGTM when the test failures are fixed.

watson:
I couldn't reproduce the bug locally. I think it's just an unrelated fluke race condition. I restarted the test for now - let's see how it goes

watson:
The codecov failures or not really failures, but just an issue with how we calculate it. I'll merge as is

watson:
Released in v7.2.0

PR #57, opened 2019-04-16
feat: support global labels

Qard:
This is the http client half of elastic/apm-agent-nodejs#963.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Data sent to the APM Server as part of the metadata package:
 - `kubernetesNamespace` - Kubernetes namespace
 - `kubernetesPodName` - Kubernetes pod name, if not given will be the hostname
 - `kubernetesPodUID` - Kubernetes pod id, if not given will be parsed from `/proc/self/cgroup`
+- `globalLabels` - An object of key/value pairs to use to label all data reported (only applied when using APM Server 7.1+)
 
 HTTP client configuration:
 
--- a/index.js
+++ b/index.js
@@ -107,6 +107,8 @@ function Client (opts) {
 Client.prototype.config = function (opts) {
   this._conf = Object.assign(this._conf || {}, opts, { objectMode: true })
 
+  this._conf.globalLabels = normalizeGlobalLabels(this._conf.globalLabels)
+
   const missing = requiredOpts.filter(name => !this._conf[name])
   if (missing.length > 0) throw new Error('Missing required option(s): ' + missing.join(', '))
 
@@ -474,7 +476,9 @@ function getMetadata (opts) {
       agent: {
         name: opts.agentName,
         version: opts.agentVersion
-      }
+      },
+      framework: undefined,
+      version: undefined
     },
     process: {
       pid: process.pid,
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -72,6 +72,7 @@ Data sent to the APM Server as part of the metadata package:
 - `kubernetesNamespace` - Kubernetes namespace
 - `kubernetesPodName` - Kubernetes pod name, if not given will be the hostname
 - `kubernetesPodUID` - Kubernetes pod id, if not given will be parsed from `/proc/self/cgroup`
+- `globalLabels` - An object of key/value pairs to use to label all data reported
```
watson: We should add a note about that this requires APM Server 7.1 in order to be picked up

index.js
```diff
@@ -554,3 +561,14 @@ function destroyStream (stream) {
     else if (typeof stream.emit === 'function') stream.emit('close')
   }
 }
+
+function normalizeGlobalLabels (labels) {
+  const result = {}
+  if (!labels) return result
+
+  for (let key of Object.keys(labels)) {
+    result[key] = labels[key].toString()
```
watson: We should verify if we're allowed to support numbers and booleans, and not only strings

index.js
```diff
@@ -485,8 +489,11 @@ function getMetadata (opts) {
     system: {
       hostname: opts.hostname,
       architecture: process.arch,
-      platform: process.platform
-    }
+      platform: process.platform,
+      container: undefined,
+      kubernetes: undefined
+    },
+    labels: opts.globalLabels
```
watson: I couldn't find any documentation for this, so just double check that the root of the payload is the right place to add the new `labels` property 😅

index.js
```diff
@@ -554,3 +561,14 @@ function destroyStream (stream) {
     else if (typeof stream.emit === 'function') stream.emit('close')
   }
 }
+
+function normalizeGlobalLabels (labels) {
+  const result = {}
+  if (!labels) return result
```
watson: If you move this up as the first line and return undefined, then it will not take up room on the metadata payload

index.js
Qard: I checked the apm-server PR and the spec files. This is the right place for it. :)

README.md
Qard: fixed

index.js
Qard: fixed

index.js
vigneshshanmugam: From 6.7 APM server accepts it - https://github.com/elastic/apm-server/pull/1712, But we are doing the same `toString` because we support 6.5+ versions.

Qard: In that case, I would consider it a breaking change to make this support numbers and booleans, so we should do that in another PR targeted at landing in the next major.

Qard: Actually, it would just get fully ignored on pre-6.7 versions because global labels are on metadata messages only, right? So it actually _wouldn't_ be a backwards-compatibility concern to send the data up with numbers and booleans. 🤔

Qard: I've changed it to allow number and boolean types.

index.js
```diff
@@ -474,7 +476,9 @@ function getMetadata (opts) {
       agent: {
         name: opts.agentName,
         version: opts.agentVersion
-      }
+      },
+      framework: undefined,
+      version: undefined
```
vigneshshanmugam: shouldn't this be opts.serviceVersion ?

Qard: Sorry, it's not immediately obvious why this is set to `undefined` here. It _is_ set to `opts.serviceVersion` later, if it has been provided. The reason for setting to `undefined` _here_ rather than just omitting it completely is a performance optimization related to the [hidden classes](https://richardartoul.github.io/jekyll/update/2015/04/26/hidden-classes.html) problem.

Qard:
Test failures are unrelated. Going to land anyway.

PR #61, opened 2019-06-17
chore(package): bump all devDependencies

watson:


watson:
Travis tests are unrelated: It's an issue that happens 9 out of 10 times on Node.js 6. There's no need to spend time fixing it as we're dropping support for Node.js 6 ASAP.

Issue #64, opened 2019-06-28
Enable CI pipeline in the apm-ci


v1v:
Let's enable the CI pipeline in the `apm-ci` instance

PR #65, opened 2019-06-28
ci(jenkins): enable jenkins multibranch pipeline

v1v:
Enable Jenkins Pipeline in http://apm-ci.elastic.co

Closes https://github.com/elastic/apm-nodejs-http-client/issues/64

## Highlights
- This particular PR is a subset of https://github.com/elastic/apm-agent-nodejs/pull/1025
- It does use docker and docker-compose to be able to select what nodejs versions are required to be tested out.

## How to test locally using the same CI setup?
- It does require `docker` and `docker-compose`.
- To test a specific nodejs version then `.ci/scripts/test.sh <NODE_VERSION>` where `<NODE_VERSION>` matches any of the current released nodejs versions. For instance: `.ci/scripts/test.sh 11`
- To generate the `junit` reports then `docker run --rm -ti -v `pwd`:/app -w /app node:12 .ci/scripts/convert_tap_to_junit.sh` and open `target/junit-test-suite-output.tap-report.xml`

## Tasks
- [x] Enable JJBB.
- [x] Enable [webhook](https://github.com/elastic/apm-nodejs-http-client/settings/hooks/120803027).
- [x] Create a vault token for the codecov service.
- [x] Enable GH service accounts.
- [x] Document what are the requirements to run this locally.


.ci/scripts/docker-test.sh
```diff
@@ -0,0 +1,15 @@
...
+id
+node --version
+npm --version
+npm install
+npm list
+
+# Let's store the test output in the target folder
+TARGET_FOLDER=target
+mkdir -p target
+nyc npm test | tee ${TARGET_FOLDER}/test-suite-output.tap || true
```
v1v: Let's fail the build as we do with the nodejs agent:
- https://github.com/elastic/apm-agent-nodejs/blob/master/.ci/scripts/docker-test.sh#L17
```suggestion
nyc npm test | tee ${TARGET_FOLDER}/test-suite-output.tap
```

.ci/.jenkins_nodejs.yml
```diff
@@ -0,0 +1,11 @@
+NODEJS_VERSION:
+  - "12"
+  - "12.0"
```
watson: While I guess it's technically going to give us more coverage, we don't currently test the http client on the `.0` versions of Node.js. So feel free to remove those from this file

kuisathaverat:
@watson If the build does not exist in Jenkins we cannot manage it, it would be isolated from the rest and could not be integrated with the rest of processes we are automated. IMHO isn't an option that it isn't in Jenkins.

.ci/.jenkins_nodejs.yml
Qard: I think the *.0 versions are mostly just an advantage with testing instrumentation is applied at the right place, but _may_ sometimes help catching use of new language features added mid-way through a version. 🤔

v1v: I've just found I resolved this particular conversation, I removed those `.0` versions, so happy to put it back if it adds any value in this particular case

.ci/.jenkins_nodejs.yml
```diff
@@ -0,0 +1,7 @@
+NODEJS_VERSION:
+  - "12"
+  - "11"
+  - "10"
+  - "8"
+  - "8.1"
```
Qard: @watson We don't need 8.1 either, do we? We just have that in the agent for async_hooks vs not, which isn't used in the http client at all, right?

.ci/Jenkinsfile
```diff
@@ -0,0 +1,96 @@
...
+  environment {
+    REPO = 'apm-nodejs-http-client'
+    BASE_DIR = "src/github.com/elastic/${env.REPO}"
+    PIPELINE_LOG_LEVEL='INFO'
+    NOTIFY_TO = credentials('notify-to')
+    JOB_GCS_BUCKET = credentials('gcs-bucket')
+    CODECOV_SECRET = "secret/apm-team/ci/${env.REPO}-codecov"
+  }
+  options {
+    timeout(time: 3, unit: 'HOURS')
```
Qard: I don't think the timeout needs to be anywhere near this high for the http client.

.ci/docker/node-container/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN ( [ "${NODE_VERSION%%.*}" -le 8  ] \
+  && echo "Node.js ${NODE_VERSION} - Manual install npm" \
+  && mkdir -p /npm/node_modules \
+  && npm install npm \
+  && rm /usr/local/bin/npm \
+  && ln -s /npm/node_modules/npm/bin/npm-cli.js /usr/local/bin/npm ) || exit 0
+
+RUN ( [ "${NODE_VERSION%%.*}" -gt 8  ] \
+  && echo "Node.js ${NODE_VERSION} - upgrade npm" \
+  && npm install -g npm@latest) || exit 0
```
Qard: Reminder we're planning on dropping pre-8.x versions soon and therefore should remember to clean this up right away, if we land this first.

.ci/.jenkins_nodejs.yml
watson: I just mentioned it because we didn't already do this, so if y'all were trying to replicate our existing Travis setup, then this wasn't needed. And if we want to be economical about how many resources we spin up, then this was a place where I think we could save a little without worrying too much about it. But if we want to add all the `x.0` versions I'm of course fine with that as well.

.ci/.jenkins_nodejs.yml
watson: Nope. As I just mentioned in https://github.com/elastic/apm-nodejs-http-client/pull/65#discussion_r301010991, then I think we either should have all or none. So if we want `8.1`, then we should also include `6.0`, `10.0`, `11.0` and `12.0` - or we should just remove `8.1`.

.ci/Jenkinsfile
watson: 10 minutes should be enough. Currently our tests take less than 2 minutes

.ci/docker/node-container/Dockerfile
watson: I think this was carried over from the agent as well. We don't need to upgrade npm in any other repo than on the agent repo.

Qard: Ah, right. In that case, we should just remove the upgrade altogether, since that would more closely match the configuration of most users.

.ci/Jenkinsfile
v1v: Sure

.ci/docker/node-container/Dockerfile
v1v: Alright, I'll remove the `npm install`. In fact, this particular implementation was taken from the agent, as Thomas mentioned.

.ci/.jenkins_nodejs.yml
v1v: I'll remove `8.1` then

v1v:
All the comments have been addressed and I guess we can potentially merge this PR and iterate if required with some follow-ups. What do you think @watson  and @Qard ard?

PR #70, opened 2019-07-29
chore: drop Node.js 6.x and 11.x

Qard:
This drops Node.js 6.x from the test matrix which should be considered a breaking change and should therefore be followed by a major bump.


watson:
Let's remove Node.js 11 as well in the same go

Qard:
Done. 👍

watson:
Ups, I forgot you already had this PR open. I guess you want to close it in favor of #81, right? Sorry about that 😬

PR #75, opened 2019-08-22
ci(jenkins): avoid builds in the master worker

v1v:
## Highlights
- `master` worker could be used but causes bottlenecks in some cases.
- This will allow to use of another worker and get rid of any potential bottlenecks when the build queue is massive.


watson:
Just ignore the failing Travis test

PR #76, opened 2019-08-29
feat: added nodejs runtime name and version to user-agent string

### LINKED ISSUE #72: Add Node.js runtime name + version to user-agent string
We should add `node/{version}` to the end of the user-agent string, e.g `node/12.4.1`:

```
${process.release.name}/${process.versions.node}
```

alemagio:
This is one of my first PR so any feedback is really appreciated

Fixes #72

### DIFF
--- a/index.js
+++ b/index.js
@@ -529,7 +529,7 @@ function getHeaders (opts) {
   const headers = {}
   if (opts.secretToken) headers['Authorization'] = 'Bearer ' + opts.secretToken
   headers['Accept'] = 'application/json'
-  headers['User-Agent'] = opts.userAgent + ' ' + pkg.name + '/' + pkg.version
+  headers['User-Agent'] = `${opts.userAgent} ${pkg.name}/${pkg.version}/${process.release.name}/${process.versions.node}`
   return Object.assign(headers, opts.headers)
 }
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -529,7 +529,7 @@ function getHeaders (opts) {
   const headers = {}
   if (opts.secretToken) headers['Authorization'] = 'Bearer ' + opts.secretToken
   headers['Accept'] = 'application/json'
-  headers['User-Agent'] = opts.userAgent + ' ' + pkg.name + '/' + pkg.version
+  headers['User-Agent'] = `${opts.userAgent} ${pkg.name}/${pkg.version}/${process.release.name}/${process.versions.node}`
```
Qard: There should be a space before `${process.release.name}`, not a slash. We want the user agent to be something like `apm-nodejs-agent/1.2.3 apm-nodejs-http-client/1.2.3 node/1.2.3`. The first part comes from the agent, the second part is the `${pkg.name}/${pkg.version}` part, and we want to add is another name/version pair for the Node.js runtime.

alemagio: Thanks for the explanation
I hope is ok now :)

watson:
jenkins run the tests please

watson:
Looks great @alemagio! Congratulations on your first PR 💯 

The reason why the Travis tests fail on Node.js 6 is not related to your code, so all is good 👍

alemagio:
> Looks great @alemagio! Congratulations on your first PR 100
> 
> The reason why the Travis tests fail on Node.js 6 is not related to your code, so all is good +1

Yeah I tried node 6 on my machine and all tests passed
Thanks guys

PR #78, opened 2019-09-02
fix: race-condition that could leave stream in corked state

watson:
If running Node.js 10.2.0 or newer, a race condition would occur in the following scenarios that would leave the stream permanently corked:

- Call `flush` while the stream is corked and `bufferWindowTime !== -1`
- Call `sendSpan`, `sendTransaction`, `sendError`, or `sendMetricSet` while stream is corked, `bufferWindowTime === -1` and `_writableState.length >= bufferWindowSize`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -310,7 +310,10 @@ Client.prototype._maybeUncork = function () {
       if (this.destroyed === false) this.uncork()
     })
 
-    if (this._corkTimer) clearTimeout(this._corkTimer)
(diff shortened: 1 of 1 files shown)


watson:
While reviewing this with @sqren we found that the stream would fix the issue by uncorking itself at some point when `bufferWindowSize` number of events had been written to the client. So this race-condition not as critical as I first thought. It just means that there will go a little more time until users see data in the APM UI

PR #81, opened 2019-09-29
chore: drop support for older versions of Node.js

watson:
BREAKING CHANGE: Only Node.js ^8.6.0, 10.x, and 12.x+ are now supported.


Qard:
Seems to be some tests still failing on `ppid`.

PR #93, opened 2019-10-25
fix(truncate): nulls should not be treated as objects

Qard:
This fixes a bug introduced in #92.


Qard:
jenkins run the tests please

PR #99, opened 2019-11-09
perf: re-use encoded metadata

### LINKED ISSUE #35: Only build metadata payload NDJSON string once
Only stringify the `metadata` object once and reuse the generated string for all future metadata events being sent.

alemagio:
Closes #35


alemagio:
Hi, is there something i need to change in order to make it pass?

PR #109, opened 2020-08-27
Cloud Metadata

JCMais:
See https://github.com/elastic/apm-agent-nodejs/issues/1815 and https://github.com/elastic/apm-agent-nodejs/pull/1816

- [x] Implement code
- [ ] Add tests

### DIFF
--- a/index.js
+++ b/index.js
@@ -71,6 +71,8 @@ function Client (opts) {
   this._corkTimer = null
   this._received = 0 // number of events given to the client for reporting
   this.sent = 0 // number of events written to the socket
+  this._hasSentMetadata = false
+  this._callsWaitingMetadata = []
   this._active = false
   this._onflushed = null
   this._transport = null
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -345,24 +347,41 @@ Client.prototype._encode = function (obj, enc) {
...
+  this._callsWaitingMetadata = []
+}
+
+Client.prototype._sendOrAddToWaitingMetadata = function(...args) {
+  if (this._hasSentMetadata) {
+    this._maybeUncork()
+    return this.write(...args)
+  } else {
+    if (this._callsWaitingMetadata.length === 0) {
+      this._chopper.write('')
```
JCMais, line 364: This was basically to trigger the metadata to be written first, as it would only ever go through this code path if this is the request that would cause a new stream to be created (`this._hasSentMetadata` is `true` and `this._callsWaitingMetadata` is empty).

We cannot use cork for that as it's not possible to force write some data to the stream while it's corked.

If there is a better way to do this, please let me know.


Return value on sendSpan / sendTransaction is always true when waiting for the metadata to be sent, this has no implications on the current code base as those return values are ignored.


Called write('') to trigger streamchopper

index.js
```diff
@@ -583,7 +621,12 @@ function getMetadata (opts) {
       container: undefined,
       kubernetes: undefined
     },
-    labels: opts.globalLabels
+    labels: opts.globalLabels,
+    cloud: opts.cloud
```
JCMais, line 625: This would allow to pass the cloud metadata directly if there is ever a need for that.

index.js
```diff
@@ -345,24 +347,41 @@ Client.prototype._encode = function (obj, enc) {
...
+Client.prototype._sendOrAddToWaitingMetadata = function(...args) {
+  if (this._hasSentMetadata) {
+    this._maybeUncork()
+    return this.write(...args)
+  } else {
+    if (this._callsWaitingMetadata.length === 0) {
+      this._chopper.write('')
+    }
+    this._callsWaitingMetadata.push(args)
+    return true
```
JCMais, line 367: This would make the return value for `sendSpan`, `sendTransaction`, etc to always be true when waiting for the metadata to be written, this has no implications on the current code base as those return values are ignored.

astorm:
closing in favor of https://github.com/elastic/apm-nodejs-http-client/pulls -- thanks again for the prototyping @JCMais!

PR #110, opened 2020-09-14
Support api keys for authentication

gergof:
This would allow the apm-agent-nodejs to be used with api keys not only secret tokens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ HTTP client configuration:
 - `userAgent` - (required) The HTTP user agent that your module should
   identify it self as
 - `secretToken` - The Elastic APM intake API secret token
+- `apiKey` - Elastic APM API key
 - `serverUrl` - The APM Server URL (default: `http://localhost:8200`)
 - `headers` - An object containing extra HTTP headers that should be
   used when making HTTP requests to he APM Server
--- a/index.js
+++ b/index.js
@@ -545,6 +545,7 @@ function getBasicRequestOptions (method, defaultPath, headers, opts, agent) {
 function getHeaders (opts) {
   const headers = {}
   if (opts.secretToken) headers.Authorization = 'Bearer ' + opts.secretToken
+  if (opts.apiKey) headers.Authorization = 'ApiKey ' + opts.apiKey
   headers.Accept = 'application/json'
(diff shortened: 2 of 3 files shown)


gergof:
>  Author of the following commits did not sign a [Contributor Agreement](https://www.elastic.co/contributor-agreement):
> [1c25054](https://github.com/elastic/apm-nodejs-http-client/commit/1c25054f5335e2d91bbba87348595bc851cbafc3)
> 
> Please, read and sign the above mentioned agreement if you want to contribute to this project

Done

astorm:
jenkins run the tests

astorm:
jenkins run the tests please

README.md
```diff
@@ -82,6 +82,7 @@ HTTP client configuration:
 - `userAgent` - (required) The HTTP user agent that your module should
   identify it self as
 - `secretToken` - The Elastic APM intake API secret token
+- `apiKey` - Elastic APM API key
 - `serverUrl` - The APM Server URL (default: `http://localhost:8200`)
```
astorm, line 86: I see here we've added the `apiKey` key to the list of configuration values. 👍

index.js
```diff
@@ -545,6 +545,7 @@ function getBasicRequestOptions (method, defaultPath, headers, opts, agent) {
 function getHeaders (opts) {
   const headers = {}
   if (opts.secretToken) headers.Authorization = 'Bearer ' + opts.secretToken
+  if (opts.apiKey) headers.Authorization = 'ApiKey ' + opts.apiKey
   headers.Accept = 'application/json'
```
astorm, line 549: I see here that when we're generating headers to send to the HTTP server that we consult the configuration for an `apiKey` value.  If this value is present we prefer its configuration over the secret token authorization as a value for the authorization header. This value is passed "raw" from the configuration with no extra encoding -- this means either the base64 encoding happens in the agent proper, or that we're expecting users to provide us with a base64 encoded key. 👍

test/config.js
```diff
@@ -75,6 +75,24 @@ test('no secretToken', function (t) {
   })
 })
 
+test('has apiKey', function (t) {
```
astorm, line 78: I see here we've created a mocked APMServer and use that mock to ensure the configuration value end up in the Authorization header.  This test matches the pattern used for other values. 👍

astorm (approved):
This looks like the necessary pre-work to pass on the authorization value set in the agent property.  Approving.

PR #114, opened 2020-10-19
feat: ensure podId conforms to elastic naming conventions

astorm:
Closes: https://github.com/elastic/apm-agent-nodejs/issues/1821

This PR ensure the podID we extract from cgroups info [conforms to the following spec change](https://github.com/elastic/apm/pull/344/files).

This change is made here instead of the `container-info` package.  The `container-info` package is in use outside of elastic, and changing it's behavior would impact multiple parties (no _ to - conversion).  Instead we rely on this package to grab the container information, and then make any additional string formatting changes needed by elastic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,3 +307,4 @@ subsequent calls to `sendSpan()`, `sendTransaction()`, `sendError()`,
 ## License
 
 [MIT](LICENSE)
+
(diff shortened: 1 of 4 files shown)


astorm:
closing in favor of a new PR

PR #116, opened 2020-10-28
feat: ensure podId conforms to naming conventions

astorm:
Moves https://github.com/elastic/apm-agent-nodejs/issues/1821 forward.

This PR ensure the podID we extract from cgroups info [conforms to the following spec change](https://github.com/elastic/apm/pull/344/files).

This change is made here instead of the `container-info` package.  The `container-info` package is in used outside of elastic, and changing its behavior would impact multiple parties.  Instead we rely on this package to grab the container information, and then add a new module for making any additional changes required locally. 

Once this lands we'll need to do a release of `apm-nodejs-http-client`.  So long as the semver bump is patch or minor, we shouldn't need a version bump in the agent proper (it's [using `^9.4`](https://github.com/elastic/apm-agent-nodejs/blob/3a845efe2fc3278135a0ab8ac9a4aa5f832605f4/package.json#L85)).  Agent users would be able to get the latest client with an `npm update`.

### DIFF
--- a/lib/container-info.js
+++ b/lib/container-info.js
@@ -6,6 +6,8 @@ const getContainerInfo = require('container-info')
  * and apply additional naming rules
  *
  * Injectable getInfo for testing purposes
+ *
+ * https://github.com/elastic/apm/blob/master/specs/agents/metadata.md#containerkubernetes-metadata
  */
 module.exports = (getInfo = getContainerInfo) => {
   const info = getInfo.sync()
--- a/lib/container-info.js
+++ b/lib/container-info.js
@@ -1,0 +1,18 @@
+const getContainerInfo = require('container-info')
+
+/**
+ * Function that uses `container-info`
+ * package to extract container information
+ * and apply additional naming rules
+ *
+ * Injectable getInfo for testing purposes
+ */
+module.exports = (getInfo = getContainerInfo) => {
+  const info = getInfo.sync()
+  if (!info) { return info }
+
+  if (info.podId) {
+    info.podId = info.podId.replace(/_/g, '-')
+  }
+  return info
+}
--- a/test/k8s.js
+++ b/test/k8s.js
@@ -277,7 +277,6 @@ test('Tests for ../lib/container-info ', function (t) {
   for (const [, fixture] of fixtures.entries()) {
     const mock = createMockForFixtureString(fixture.source)
     const info = containerInfo(mock)
-    // console.log(info)
     t.equals(info.podId, fixture.expectedPodId, 'expected pod ID returned')
   }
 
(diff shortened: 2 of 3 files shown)


test/k8s.js
```diff
@@ -210,3 +212,74 @@ test('all except kubernetesPodUID', function (t) {
...
+    },
+    {
+      source: '1:name=systemd:/kubepods/besteffort/pod74c13223-5a00-11e9-b385-42010a80018d/34dc0b5e626f2c5c4c5170e34b10e7654ce36f0fcd532739f4445baabea03376',
+      expectedPodId: '74c13223-5a00-11e9-b385-42010a80018d'
+    }
+  ]
+  for (const [, fixture] of fixtures.entries()) {
+    const mock = createMockForFixtureString(fixture.source)
+    const info = containerInfo(mock)
+    // console.log(info)
```
vigneshshanmugam, line 280: nit: remove console.log

vigneshshanmugam (approved):
Looks good.

jahtalab (approved):

test/k8s.js
astorm: 🤦 -- I need to dig up my linting rule for catching these :)

test/k8s.js
```diff
@@ -210,3 +212,74 @@ test('all except kubernetesPodUID', function (t) {
     client.flush()
   })
 })
+
+test('Tests for ../lib/container-info ', function (t) {
```
trentm, line 216: nit: Drop the trailing space on the test name.

trentm (commented):
Isn't this missing:

```
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ const { URL } = require('url')
 const zlib = require('zlib')
 const querystring = require('querystring')
 const Writable = require('readable-stream').Writable
-const getContainerInfo = require('container-info')
+const getContainerInfo = require('./lib/container-info')
 const pump = require('pump')
 const eos = require('end-of-stream')
 const streamToBuffer = require('fast-stream-to-buffer')
@@ -26,7 +26,7 @@ const requiredOpts = [
   'userAgent'
 ]

-const containerInfo = getContainerInfo.sync()
+const containerInfo = getContainerInfo()
```

lib/container-info.js
```diff
@@ -0,0 +1,18 @@
+const getContainerInfo = require('container-info')
+
+/**
+ * Function that uses `container-info`
+ * package to extract container information
+ * and apply additional naming rules
```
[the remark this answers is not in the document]
astorm: Good idea -- link added.

test/k8s.js
astorm: rid applied.

trentm (approved):

PR #117, opened 2020-11-08
Added support for custom http.Agent. Fixes #91

### LINKED ISSUE #91: HTTP proxy support?
Hey guys,

Sorry if this was asked before, but does this client (and more generally the NodeJS APM client) support a HTTP(s) proxy? I've quickly scrolled through the docs and I couldn't find a way to configure my proxy... Is there some undocumented, automagic environment variable ?

Regards,
David

ddolcimascolo:
Hi guys!

Following the discussion in elastic/apm-nodejs-http-client#91 I added support for passing in custom http.Agent instances, in order to be able to traverse HTTP proxies when the APM server you're connecting to is after such a proxy.
My ultimate goal is to make https://github.com/elastic/apm-agent-nodejs support the new setting, any advice on whether I should contribute something there also is welcome :+1: 

Thanks,
David

### DIFF
--- a/index.js
+++ b/index.js
@@ -73,25 +73,9 @@ function Client (opts) {
   this.sent = 0 // number of events written to the socket
   this._active = false
   this._onflushed = null
-  this._transport = null
   this._configTimer = null
   this._encodedMetadata = null
 
-  switch (this._conf.serverUrl.protocol.slice(0, -1)) { // 'http:' => 'http'
-    case 'http': {
-      this._transport = require('http')
-      break
-    }
-    case 'https': {
-      this._transport = require('https')
-      break
-    }
-    default: {
-      throw new Error('Unknown protocol ' + this._conf.serverUrl.protocol.slice(0, -1))
-    }
-  }
-
-  this._agent = new this._transport.Agent(this._conf)
   this._chopper = new StreamChopper({
     size: this._conf.size,
     time: this._conf.time,
@@ -137,6 +121,22 @@ Client.prototype.config = function (opts) {
   // process
   this._conf.serverUrl = new URL(this._conf.serverUrl)
 
+  switch (this._conf.serverUrl.protocol.slice(0, -1)) { // 'http:' => 'http'
+    case 'http': {
+      this._transport = require('http')
+      break
+    }
+    case 'https': {
+      this._transport = require('https')
+      break
+    }
+    default: {
+      throw new Error('Unknown protocol ' + this._conf.serverUrl.protocol.slice(0, -1))
+    }
+  }
+
+  this._agent = this._conf.agent || new this._transport.Agent(this._conf)
+
   if (containerInfo) {
     if (!this._conf.containerId && containerInfo.containerId) {
       this._conf.containerId = containerInfo.containerId
@@ -162,11 +162,13 @@ Client.prototype._pollConfig = function () {
     opts.headers['If-None-Match'] = this._conf.lastConfigEtag
   }
 
+  const onError = err => {
+    this._scheduleNextConfigPoll()
+    this.emit('request-error', err)
+  }
+
   const req = this._transport.get(opts, res => {
-    res.on('error', err => {
-      // Not sure this event can ever be emitted, but just in case
-      res.destroy(err)
-    })
+    res.on('error', onError) // Not sure this event can ever be emitted, but just in case
 
     this._scheduleNextConfigPoll(getMaxAge(res))
 
@@ -180,7 +182,7 @@ Client.prototype._pollConfig = function () {
     }
 
     streamToBuffer(res, (err, buf) => {
-      if (err) return res.destroy(err)
+      if (err) return onError(err)
 
       if (res.statusCode === 200) {
         // 200: New config available (or no config for the given service.name / service.environment)
@@ -190,18 +192,15 @@ Client.prototype._pollConfig = function () {
         try {
           this.emit('config', JSON.parse(buf))
         } catch (e) {
-          res.destroy(e)
+          onError(e)
         }
       } else {
-        res.destroy(processConfigErrorResponse(res, buf))
+        onError(processConfigErrorResponse(res, buf))
       }
     })
   })
 
-  req.on('error', err => {
-    this._scheduleNextConfigPoll()
-    this.emit('request-error', err)
-  })
+  req.on('error', onError)
 }
 
 Client.prototype._scheduleNextConfigPoll = function (seconds) {
--- a/test/central-config.js
+++ b/test/central-config.js
@@ -87,7 +87,6 @@ test('polling', function (t) {
         t.end()
         res.writeHead(404) // end nicely so we don't get a request-error
         res.end()
-        client.destroy()
         server.close()
         break
       default:
--- a/README.md
+++ b/README.md
@@ -104,6 +104,9 @@ HTTP client configuration:
   `Infinity`)
 - `maxFreeSockets` - Maximum number of sockets to leave open in a free
   state. Only relevant if `keepAlive` is set to `true` (default: `256`)
+- `agent` - Custom [http.Agent](https://nodejs.org/api/http.html#http_class_http_agent).
+  Typical use case is to swap the default Agent with one compatible with a
+  HTTP proxy, such as [https-proxy-agent](https://github.com/TooTallNate/node-https-proxy-agent)
 
 APM Agent Configuration via Kibana:
 
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   },
   "devDependencies": {
     "codecov": "^3.6.1",
+    "https-proxy-agent": "^5.0.0",
     "ndjson": "^1.5.0",
     "nyc": "^14.1.1",
     "semver": "^6.3.0",
--- a/test/agent.js
+++ b/test/agent.js
@@ -1,0 +1,55 @@
+'use strict'
+
+const test = require('tape')
+const utils = require('./lib/utils')
+const HttpsProxyAgent = require('https-proxy-agent')
+const ProxyServer = require('./helpers/ProxyServer')
+
+test('With a custom http.Agent and central config disabled', function (t) {
+  t.plan(1)
+
+  const proxy = new ProxyServer()
+  const server = utils.APMServer(
+    function (req, res) {
+      req.resume()
+      req.on('end', function () {
+        res.end()
+        t.deepEqual([`localhost:${server.address().port}`], proxy.requests, 'Client can traverse a HTTP proxy')
+        server.close()
+        proxy.close()
+        t.end()
+      })
+    })
+
+  proxy.listen(() => {
+    server.client(
+      { agent: new HttpsProxyAgent({ port: proxy.port() }) },
+      (client) => {
+        client.sendTransaction({ foo: 42 })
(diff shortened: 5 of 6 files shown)


astorm:
@ddolcimascolo Hello!  Thanks for the PR, this looks like it would be a welcome addition to the agent.  

We notice that the TravisCI tests [are failing here](https://travis-ci.org/github/elastic/apm-nodejs-http-client/builds/743509682) with these changes. 

We're seeing similar failures when we run `npm run test` locally.

Do you have any thoughts on why they might be failing and what you'd need to do to fix 'em?

ddolcimascolo:
Hi @astorm 

Thx for taking time to check this PR! I'll look into the tests failures asap and post back here. I moved a bit of configuration code in order to be able to init the custom Agent correctly, I might have overlooked something.

David

ddolcimascolo:
@astorm I believe tests pass now :)

astorm:
jenkins run the tests please

astorm:
jenkins run the tests please

ddolcimascolo:
Hi guys, I just rebased on latest master and I believe tests are passing for quite a few weeks now. Anything else you need from me before this can be merged? For the record, my goal is to make the actual nodejs APM agent compatible with this, if you want me to make a pull request there, just ask :+1: 

Cheers,
David

astorm:
jenkins run the tests

astorm:
@ddolcimascolo Thanks for checking in and sorry for losing you in the noise -- the squeaky wheel definitely gets the grease from a busy maintainer team :) 

I just checked out your branch and attempted to run the tests.  I saw two failures -- and that seems to line up with what our automated GitHub CI is seeing as well.


    $ npm run test

    //...

    not ok 1226 plan != count
      ---
        operator: fail
        expected: 59
        actual:   50
        at: Server.<anonymous> (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/test/central-config.js:87:11)
        stack: |-
          Error: plan != count
              at Test.assert [as _assert] (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/node_modules/tape/lib/test.js:228:54)
              at Test.bound [as _assert] (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/node_modules/tape/lib/test.js:80:32)
              at Test.fail (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/node_modules/tape/lib/test.js:322:10)
              at Test.bound [as fail] (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/node_modules/tape/lib/test.js:80:32)
              at Test._end (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/node_modules/tape/lib/test.js:173:14)
              at Test.bound [as _end] (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/node_modules/tape/lib/test.js:80:32)
              at Test.end (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/node_modules/tape/lib/test.js:157:10)
              at Test.bound [as end] (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/node_modules/tape/lib/test.js:80:32)
              at Server.<anonymous> (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/test/central-config.js:87:11)
              at Server.emit (events.js:315:20)
      ...
      
    //...

    not ok 1241 Error: socket hang up
      ---
        operator: error
        expected: |-
          undefined
        actual: |-
          { [Error: socket hang up] code: 'ECONNRESET' }
        at: Client.<anonymous> (/Users/astorm/Documents/github/elastic/apm-nodejs-http-client/test/central-config.js:116:11)
        stack: |-
          Error: socket hang up
              at connResetException (internal/errors.js:617:14)
              at Socket.socketCloseListener (_http_client.js:443:25)
              at Socket.emit (events.js:327:22)
              at TCP.<anonymous> (net.js:673:12)
      ...
  
We'll want/need these to pass before we can review and merge. 

If you need any help running the tests (or with anything) let us know and we'll respond as best we're able. 

Thanks again for taking the time to to make these PRs, they are appreciated.

ddolcimascolo:
Ah yes sorry, tests were green before the rebase. I didn't pay attention. I'll fix those

astorm:
jenkins the the tests

astorm:
jenkins run the tests please

astorm (changes requested):
@ddolcimascolo -- thanks for the merge tidying.

In reviewing this work we discovered a a pre-existing inconsistency in our current use of  http-agents in our http connections.  This isn't exactly _your_ issue but may impact what you're trying to do here. 

Currently, in a `Client`'s constructor function, we 

1. Call `this.config()`: https://github.com/elastic/apm-nodejs-http-client/blob/164775608ced60ecddfec5223505ec2927d9bbba/index.js#L59 
2. Assign a value `this._agent` _after_ the call to `this.config`: https://github.com/elastic/apm-nodejs-http-client/blob/164775608ced60ecddfec5223505ec2927d9bbba/index.js#L94

However, -- in the `config` method itself we create our request option objects (`this._conf.requestIntake` and `this._conf.requestConfig`), which uses the value of `this._agent`: https://github.com/elastic/apm-nodejs-http-client/blob/164775608ced60ecddfec5223505ec2927d9bbba/index.js#L153
    
This means there's a brief period of time between the `Client` finishing its instantiation and the next call to `.config` where the http-agent that's set at both `this._conf.requestIntake.agent` and `this._conf.requestConfig.agent` is `undefined`.  

In turn, this means that there's a single request to the configuration endpoint that will always use an _undefined_ http-agent.  Also, it _may be possible_ (depending on the length of time it takes the configuration endpoint to respond, the behavior of the elastic-agent and (_hand wave_) async callback scheduling) that the first `onStream` callback will be called before `this._conf.requestIntake.agent` is set correctly.  

Up until now this inconsistency didn't meaningfully impact the default functionality of the elastic-agent. However -- if the intent here is to ensure the elastic-agent can work with a proxy server then this may be a problem we should address now by making sure that the right http-agent is assigned before its used in any connections.

@ddolcimascolo what do you think about the above?  Do you think we need to handle this now in order to ensure things work?

ddolcimascolo:
@astorm I totally understand and I actually spotted that while fixing the tests. I just don't have enough knowledge of the library internals, yet :)

I'm not exactly in a hurry, so yeah, I think we should handle that now in order to properly support custom agents. I'll work on it ASAP, probably by the end of the week, and update the PR.

ddolcimascolo:
So... Now that the defined `Agent` is always used, the `_pollConfig` code does not work as expected because it expects the underlying socket to close (and emit an 'error' event) when `IncomingMessage.destroy()` is called on the received HTTP response. This is not the case when using keep alives, which is the case by default.

I've refactored the code, tests are passing but I had to remove a `client.destroy()` call, in tests, otherwise I always get a ECONNRESET error that I couldn't really explain.

Tell me what you think !

Regards,
David

astorm:
jenkins run the tests please

astorm:
@ddolcimascolo this is still on our radar, but we're in the middle of getting some work done for a platform release so we haven't been able to give it the thorough review it deserves.  

Taking a quick look though I had two thoughts

1. As you've discovered, the client makes _two_ HTTP requests to APM Server.  One is for the configuration polling, the second is/are the requests made that stream agent data to the request object and then on the APM Server.  Because the nature of these two HTTP requests is different (the polling is a traditional quick open/close HTTP request, the other is a "kept open for 10 seconds" request) do you think we need the ability to configure an http-agent for each one?  Or is one agent configuration enough (non-leading question -- I haven't thought hard enough about this)

2. The removal of the `client.destroy()` in the tests is a little concerning.  I'd like to understand a bit more about what the connect/disconnect flow looks like that made this necessary and whether it might impact actual agent behavior. I intend to look into this but don't have a timeframe for that -- if anyone reading wants to go deep on this please accept my pre-gratitude.

ddolcimascolo:
Thx @astorm It's OK with me, I don't have any urging need for this one and I'd rather get it right even if that takes a bit longer.

index.js
```diff
@@ -190,18 +192,15 @@ Client.prototype._pollConfig = function () {
         try {
           this.emit('config', JSON.parse(buf))
         } catch (e) {
-          res.destroy(e)
+          onError(e)
         }
```
astorm, line 196: I see here we've replaced a simple `res.destroy` call with a call to `onError`.  The old code will destroy the response object and do nothing else. The new code _does not_ destroy the response object, and also schedules a new configuration poll, and also emits an additional `request-error` event.  My naive assumption is we'd want to maintain the existing agent behavior -- can you talk a little bit about what you're trying to do here so we can figure out the best way to go about it?

index.js
```diff
@@ -190,18 +192,15 @@ Client.prototype._pollConfig = function () {
...
       } else {
-        res.destroy(processConfigErrorResponse(res, buf))
+        onError(processConfigErrorResponse(res, buf))
       }
```
astorm, line 199: I see here we've replaced a simple `res.destroy` call with a call to `onError`.  The old code will destroy the response object and do nothing else. The new code _does not_ destroy the response object, and also schedules a new configuration poll, and also emits an additional `request-error` event.  My naive assumption is we'd want to maintain the existing agent behavior -- can you talk a little bit about what you're trying to do here so we can figure out the best way to go about it?

test/central-config.js
```diff
@@ -87,7 +87,6 @@ test('polling', function (t) {
         t.end()
         res.writeHead(404) // end nicely so we don't get a request-error
         res.end()
-        client.destroy()
         server.close()
```
astorm, line 90: I wasn't able to track down specifically why this line caused an error in your tests -- but it defintely seems related to losing the `destroy` calls I mentioned above.  Let's get those back in there (or talk about their intent so we can come up with a better solution) and see if this call to `destroy` is still necessary.

astorm (changes requested):

index.js
```diff
@@ -162,11 +162,13 @@ Client.prototype._pollConfig = function () {
...
+    this._scheduleNextConfigPoll()
+    this.emit('request-error', err)
+  }
+
   const req = this._transport.get(opts, res => {
-    res.on('error', err => {
-      // Not sure this event can ever be emitted, but just in case
-      res.destroy(err)
-    })
+    res.on('error', onError) // Not sure this event can ever be emitted, but just in case
```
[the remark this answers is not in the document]
ddolcimascolo: This reply is for all the comments related to that change.

This is what I was trying to explain initially, but I admit this is hard to explain... So one more try!
The old code calls `res.destroy` which will:
 - Kill the socket if keepAlives are disabled, which will fire a `error` event on the request, which will make the Client emit a `request-error` event. Yes, the `request-error` event was sent, there are tests that proves it but again, when `keepAlives` are disabled.
 - Do nothing at all when `keepAlives` are enabled, which is expected because the socket stays open for future request handling. In this case the `request-error` event logic wasn't triggered at all. That's why I'm manually calling it now
 
keepAlives are enabled by default but tests were passing because the Agent was ignored in the code, as you spotted in a previous review.

So my changes standardize the error handling, no matter if keepAlives are enabled or not, making the tests green.

I hope that my intent and the effects of the code are clearer now :)

ddolcimascolo:
Guys, I'm going to close this one. I'll try to open a fresh PR in the future

PR #123, opened 2020-12-23
s/Travis CI/GitHub Actions

trentm:
This also changes which versions of node tested:
- drop node v13 (only test with latest odd major)
- add node v14 (LTS) ~and node v15 (current latest odd)~ (node v15 testing moved to #124)

Refs: https://github.com/elastic/apm-agent-nodejs/issues/1922


trentm:
I think this may legitimately be finding an issue with node v15. I might push that out to a separate change.

trentm:
The incomplete check is the old *travis* one. That'll go away after I remove the travis webhook, modulo some caching.

astorm (approved):
👍

PR #125, opened 2021-01-05
chore: simplify client initialization code

watson:
Just a nit I found while I was browsing the code - A previous refactor had made the slice redundant.


trentm:
This was effectively done as part of #139

PR #126, opened 2021-01-05
Accept Cloud Metadata Payload

astorm:
This is the HTTP client portion of https://github.com/elastic/apm-agent-nodejs/issues/1815

This PR

1. Adds a new top level `cloudMetadata` configuration field to the configuration options
2. Passes this data on to the `cloud` key of the metadata payload that's sent to APM server
3. Implements truncation rules, following the pattern set for other metadata fields

This PR treats the client as a dumb pipe and does not make heroic efforts to validate the data, (other than truncating too long data, as #3)

Spec: https://github.com/elastic/apm/blob/3f224a9cbacb196bbb196b0868d9fa90a0bbaf6c/specs/agents/metadata.md#cloud-provider-metadata

Schema: https://github.com/elastic/apm-server/blob/master/docs/spec/v2/metadata.json

### DIFF
--- a/index.js
+++ b/index.js
@@ -621,7 +621,7 @@ function getMetadata (opts) {
   }
 
   if (opts.cloudMetadata) {
-    payload.cloud = Object.assign({},opts.cloudMetadata)
+    payload.cloud = Object.assign({}, opts.cloudMetadata)
   }
 
   return payload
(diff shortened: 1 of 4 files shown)


trentm (approved):
So is the intent to have the agent gather the cloud metadata, if any, at startup before creating the http client?

PR #131, opened 2021-01-21
internals docs -- client streams and stream corking

astorm:
This PR creates a `docs` folder for this repo and adds a few initial documents describing how the HTTP client uses Node's underlying stream APIs to receive data and send it on to the APM server.  Created at the request @felixbarny and my own internal "how does this even work demon dogs.

### DIFF
--- a/docs/internals/client-streams.md
+++ b/docs/internals/client-streams.md
@@ -55,7 +55,7 @@ Consider this simple program
     s.write('hello stream')
 ```
 
-The MyStream class allows us to create a stream object which we can
+The `MyStream` class allows us to create a stream object that we can
 write data to. The MyStream object can take action whenever data is
 written to this stream.
 
@@ -101,7 +101,7 @@ Let's look at each stream object independently.
 
 Whenever the Elastic Agent wants to send a piece of data to APM Server,
 it will first call a method on the apm-nodejs-http-client object. You
-can see an example [of that
+can see an example [of this
 here](https://github.com/elastic/apm-agent-nodejs/blob/3c2505b5f299381f09409500b4f0108dbb01ba38/lib/instrumentation/index.js#L241),
 with the agent calling the sendSpan method of the apm-nodejs-http-client
 object.
@@ -146,7 +146,7 @@ The stream-chopper
     stream-chopper will close and destroy the destination-stream object,
     and then create a new destination-stream
 
-5.  If the stream-chopper writes data to a stream passed a configured
+5.  If the stream-chopper writes data to a stream past a
     time limit, the stream-chopper will close and destroy the
     destination-stream object, and then create a new destination-stream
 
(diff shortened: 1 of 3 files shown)


docs/internals/stream-corking.md
```diff
@@ -0,0 +1,254 @@
...
+  })
+```
+
+ We also [_refuse_ to uncork the stream](https://github.com/elastic/apm-nodejs-http-client/blob/b992ef5a0a1eba3a2690cd275acb88938174c8fa/index.js#L350) until this metadata is present in the `_encodedMetadata` property.
+
+The effectively blocks data from being written to the stream chopper until these network requests finish.  Once the `_encodedMetadata` property is set, this specific instance of stream corking will not be invoked again.
+
+## Data Buffering
+
+The second way we use corking is to buffer our stream writes. Data is never immediately written to the stream chopper object.  Instead, the client stream will repeatedly
```
trentm, line 161: The PR (https://github.com/elastic/apm-nodejs-http-client/pull/13) that added this said it is to improve performance without more details. I gather it is to bias to using _writev (for which the writable stream mechanics inside node are somewhat better), rather than _write.

Perhaps worth clarifying this "to buffer our stream writes", is about buffering writes (a) on the Client stream, before (b) writing them in bulk to the StreamChopper.

docs/internals/stream-corking.md
```diff
@@ -0,0 +1,254 @@
+# Stream Corking
+
+As discussed elsewhere, the Elastic Node.js Agent's HTTP Client is a writable stream object.  One advanced streaming feature the client uses [is Node.js stream corking](https://nodejs.org/api/stream.html#stream_writable_cork).
+
+We use stream corking to solve two independent challenges.  The first is to delay writes to the downstream stream chopper object until the client is 100% ready to send data.  The second is to prevent what Node.js calls "head-of-line blocking" issues, when the writing of individual objects downstream can creating blocking code situations.
```
trentm, line 5: AFAIK this isn't "head of line blocking" https://en.wikipedia.org/wiki/Head-of-line_blocking

trentm (commented):
Some partial review and notes while learning the http client better for https://github.com/elastic/apm-nodejs-http-client/issues/136 work

docs/internals/stream-corking.md
astorm: _nod_ I was simply aping what the node docs themselves said. This will be easy enough to take out. https://nodejs.org/api/stream.html

docs/internals/client-streams.md
```diff
@@ -0,0 +1,376 @@
...
+### Stream Timeout/Closing
+
+Earlier we mentioned one feature of the stream-chopper was
+
+> If the stream-chopper writes data to a stream passed a configured
+> time limit, the stream-chopper will close and destroy the
+> destination-stream object, and then create a new destination-stream
+
+The stream-chopper has [[a time
+limit]{.ul}](https://github.com/elastic/apm-nodejs-http-client/blob/7f352b2181533435eee11d9da4d41a15ac607851/index.js#L97)
```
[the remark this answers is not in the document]
astorm: looks like some [pandoc detritus](https://pandoc.org/) from all the markdown to docx/google-docs transformations.  Should be easy enough to clean up.

PR #137, opened 2021-02-19
fix: calling client.send*() after client.destroy() should not error

trentm:
Before this change calling client.sendTransaction() et al after
client.destroy() would stage the data to be written, then set the
_corkTimer timeout, after which it would uncork the stream and attempt
to write:
    Error: Cannot call write after a stream was destroyed

### DIFF
--- a/index.js
+++ b/index.js
@@ -389,8 +389,16 @@ Client.prototype._encode = function (obj, enc) {
   return ndjson.serialize(out)
 }
 
+// With the cork/uncork handling on this stream, `this.write`ing on this
+// stream when already destroyed will lead to:
+//    Error: Cannot call write after a stream was destroyed
+// when the `_corkTimer` expires.
+Client.prototype._isUnsafeToWrite = function () {
+  return this.destroyed
+}
+
 Client.prototype.sendSpan = function (span, cb) {
-  if (this.destroyed) {
+  if (this._isUnsafeToWrite()) {
     return
   }
   this._maybeCork()
@@ -398,7 +406,7 @@ Client.prototype.sendSpan = function (span, cb) {
 }
 
 Client.prototype.sendTransaction = function (transaction, cb) {
-  if (this.destroyed) {
+  if (this._isUnsafeToWrite()) {
     return
   }
   this._maybeCork()
@@ -406,7 +414,7 @@ Client.prototype.sendTransaction = function (transaction, cb) {
 }
 
 Client.prototype.sendError = function (error, cb) {
-  if (this.destroyed) {
+  if (this._isUnsafeToWrite()) {
     return
   }
   this._maybeCork()
@@ -414,7 +422,7 @@ Client.prototype.sendError = function (error, cb) {
 }
 
 Client.prototype.sendMetricSet = function (metricset, cb) {
-  if (this.destroyed) {
+  if (this._isUnsafeToWrite()) {
     return
   }
   this._maybeCork()
--- a/index.js
+++ b/index.js
@@ -390,21 +390,33 @@ Client.prototype._encode = function (obj, enc) {
 }
 
 Client.prototype.sendSpan = function (span, cb) {
+  if (this.destroyed) {
+    return
+  }
   this._maybeCork()
   return this.write({ span }, Client.encoding.SPAN, cb)
 }
 
 Client.prototype.sendTransaction = function (transaction, cb) {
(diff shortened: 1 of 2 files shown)


trentm:
Here is a "bar.js" that shows the issue:

```js
var Client = require('./')

var client = new Client({
  agentName: 'myname',
  agentVersion: '1.0.0',
  serviceName: 'myService',
  userAgent: 'myUserAgent',
})

client.on('cloud-metadata', function () {
  process.nextTick(function () {
    client.destroy()
    client.sendSpan({a: 'span'})
  })
})

setInterval(function () {
  console.log('still hanging out...')
}, 3000)
```

Running that (with some `console.warn('XXX`s mixed in):

```
% node bar.js
XXX cork now in ctor
XXX uncork now in _maybeUncork
XXX _destroy
XXX sendSpan
XXX _maybeCork 0
XXX cork now in _maybeCork
XXX setTimeout _corkTimer in _maybeCork
XXX uncork now in _corkTimer
events.js:174
      throw er; // Unhandled 'error' event
      ^

Error: Cannot call write after a stream was destroyed
    at doWrite (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/_stream_writable.js:409:38)
    at clearBuffer (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/_stream_writable.js:526:7)
    at Client.Writable.uncork (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/_stream_writable.js:321:94)
    at Timeout._corkTimer.setTimeout [as _onTimeout] (/Users/trentm/el/apm-nodejs-http-client2/index.js:347:14)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)
Emitted 'error' event at:
    at errorOrDestroy (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/internal/streams/destroy.js:98:101)
    at onwriteError (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/_stream_writable.js:424:5)
    at onwrite (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/_stream_writable.js:450:11)
    at WritableState.onwrite (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/_stream_writable.js:160:5)
    at doWrite (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/_stream_writable.js:409:30)
    at clearBuffer (/Users/trentm/el/apm-nodejs-http-client2/node_modules/readable-stream/lib/_stream_writable.js:526:7)
    [... lines matching original stack trace ...]
    at tryOnTimeout (timers.js:300:5)
```

index.js
```diff
@@ -390,21 +390,33 @@ Client.prototype._encode = function (obj, enc) {
...
 Client.prototype.sendError = function (error, cb) {
+  if (this.destroyed) {
+    return
+  }
   this._maybeCork()
   return this.write({ error }, Client.encoding.ERROR, cb)
 }
 
 Client.prototype.sendMetricSet = function (metricset, cb) {
+  if (this.destroyed) {
```
astorm, line 417: nit: Would we be better off with a helper method for these checks since they're the same check for the same reason? -- Something like `Client.prototype._isSafeToWrite`?

astorm (approved):
Approving -- with the small caveat that we're continuing down a road where individual methods in the client are becoming _harder_ to reason about without a full understanding the whole. This might be another small tick in the "reconsider our streamlining logic in the wake of modern node features" column.  That said the error is real and this fixes it so :+1:

index.js
trentm: Does https://github.com/elastic/apm-nodejs-http-client/pull/137/commits/1ec1efa2cf12be6750fdba709af37d96ffa01064 look reasonable?

trentm:
> continuing down a road where individual methods in the client are becoming _harder_ to reason about without a full understanding the whole.

Renamed them to "fireSpanAndForgetMaybe"? ;)

Unrelated aside: the apm-agent-nodejs never uses the `cb` argument on these methods. I think we should drop them and document the `send*()` methods as fire and forget-y.

astorm (approved):
🚢

PR #139, opened 2021-02-23
fix: ensure keep-alive HTTP(S) agent is used

trentm:
(This is a follow up on and depends on #138 going in first.)

Fix config initialization such that the keep-alive agent is used all the
time as intended. Before this change the keep-alive HTTP(S) agent would
only be used if a second call to `client.config(...)` was made. For
the Elastic APM Agent's usage of this module, that was when any of the
express, fastify, restify, hapi, or koa modules was instrumented.

## reviewer notes

- the creation of the HTTP(S) agent (this._agent) is moved into `Client.prototype.config` -- along with setting of `this._transport` for convenience.
    - As a minor side note the setting of _transport is taking the simplification from https://github.com/elastic/apm-nodejs-http-client/pull/125/files
- The calls to `Writable.call()` and `new _transport.Agent()` have been cleaned up to only pass in supported options rather than the whole `this._conf` object, to avoid possible surprise and noise (logging the Agent instance would dump the full `_conf`).
- The significant change for handling in the test suite was in "end of test" handling ensuring that the created client's keep-alive agent was destroyed. Otherwise the process would hang around until the default keep-alive socket timeout, which is 2 minutes. (Note: sometimes that "hang" would be much shorter -- on the order of a few seconds. I'm not sure what conditions led to a sometime shorter timeout on the agent closing idle sockets.)
- The client and the APM agent test suites pass with this change. (I haven't run the TAV tests.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const http = require('http')
+const https = require('https')
 const util = require('util')
 const os = require('os')
 const { URL } = require('url')
@@ -11,6 +13,7 @@ const pump = require('pump')
 const eos = require('end-of-stream')
 const streamToBuffer = require('fast-stream-to-buffer')
 const StreamChopper = require('stream-chopper')
+
 const ndjson = require('./lib/ndjson')
 const truncate = require('./lib/truncate')
 const pkg = require('./package')
@@ -56,42 +59,19 @@ Client.encoding = Object.freeze({
 function Client (opts) {
   if (!(this instanceof Client)) return new Client(opts)
 
-  this.config(opts)
-
-  Writable.call(this, this._conf)
-
-  const errorproxy = (err) => {
-    if (this.destroyed === false) this.emit('request-error', err)
-  }
-
-  const fail = () => {
-    if (this._writableState.ending === false) this.destroy()
-  }
+  Writable.call(this, { objectMode: true })
 
   this._corkTimer = null
   this._received = 0 // number of events given to the client for reporting
   this.sent = 0 // number of events written to the socket
+  this._agent = null
   this._active = false
   this._onflushed = null
   this._transport = null
   this._configTimer = null
   this._encodedMetadata = null
 
-  switch (this._conf.serverUrl.protocol.slice(0, -1)) { // 'http:' => 'http'
-    case 'http': {
-      this._transport = require('http')
-      break
-    }
-    case 'https': {
-      this._transport = require('https')
-      break
-    }
-    default: {
-      throw new Error('Unknown protocol ' + this._conf.serverUrl.protocol.slice(0, -1))
-    }
-  }
-
-  this._agent = new this._transport.Agent(this._conf)
+  this.config(opts)
 
   // start stream in corked mode, uncork when cloud
   // metadata is fetched and assigned.  Also, the
@@ -112,6 +92,9 @@ function Client (opts) {
     this.emit('cloud-metadata', this._encodedMetadata)
   })
 
+  const errorproxy = (err) => {
+    if (this.destroyed === false) this.emit('request-error', err)
+  }
   this._chopper = new StreamChopper({
     size: this._conf.size,
     time: this._conf.time,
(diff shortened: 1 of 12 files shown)


index.js
```diff
@@ -56,42 +59,19 @@ Client.encoding = Object.freeze({
...
-
-  Writable.call(this, this._conf)
-
-  const errorproxy = (err) => {
-    if (this.destroyed === false) this.emit('request-error', err)
-  }
-
-  const fail = () => {
-    if (this._writableState.ending === false) this.destroy()
-  }
```
trentm, line 69: Reviewer note: I just moved these down in the same function beside where they are used.

index.js
```diff
@@ -56,42 +59,19 @@ Client.encoding = Object.freeze({
...
+  Writable.call(this, { objectMode: true })
```
trentm, line 62: Reviewer note: The only "Writable" option we every pass in `objectMode`, so let's only pass that. And it doesn't need to be (and shouldn't be) externally configurable.

index.js
```diff
@@ -56,42 +59,19 @@ Client.encoding = Object.freeze({
...
-      this._transport = require('https')
-      break
-    }
-    default: {
-      throw new Error('Unknown protocol ' + this._conf.serverUrl.protocol.slice(0, -1))
-    }
-  }
-
-  this._agent = new this._transport.Agent(this._conf)
+  this.config(opts)
```
trentm, line 74: Reviewer note: `this._transport` and `this._agent` are now set in `this.config()`.

test/lib/utils.js
```diff
@@ -111,7 +111,7 @@ function assertMetadata (t, obj) {
   t.ok(Array.isArray(_process.argv), 'process.title should be an array')
   t.ok(_process.argv.length >= 2, 'process.title should contain at least two elements')
   t.ok(/\/node$/.test(_process.argv[0]), `process.argv[0] should match /\\/node$/ (was: ${_process.argv[0]})`)
-  const regex = /(\/test\/(test|truncate|abort|edge-cases|lib\/unref-client)\.js|node_modules\/\.bin\/tape)$/
+  const regex = /(\/test\/.*\.js|node_modules\/\.bin\/tape)$/
```
trentm, line 114: Reviewer note: I could revert this if you like. The reason for this is to not have arbitrary test breakage when running test files directly via `node test/foo.js`. Running test files like this isn't *typical*, but to help with test debugging I temporarily flirted with switching from `tape` to `tap` (https://node-tap.org/) which has better test-ending handling and isolation and allows running test files without the "runner" CLI.

If you look at the history on this file there are a number of changes coming in augmenting this regex to handle the next failure.

test/abort.js
```diff
@@ -48,6 +48,7 @@ test('abort request if server responds early', function (t) {
         res.end()
         clearTimeout(timer)
         server.close()
+        client.destroy() // Destroy keep-alive agent.
```
trentm, line 51: Reviewer note: Every other change below (all to files under test/...), with the exception of a change in test/lib/utils.js, is about calling `client.destroy()` to properly close active handles (timers, the keep-alive net.Socket) so that the test completes with having to wait for long timeouts.

trentm (commented):
Add reviewer notes.

index.js
astorm: Just to say it out loud, this change means that _if_ someone was using the `apm-nodejs-http-client` package stand-alone to talk to APM Server that they no longer have the ability to control which configuration/opts values the underlying writable stream is instantiated with.  I don't think this is a blocker -- it's an esoteric use-case on top of an esoteric use-case and if someone runs into it they can easily rollback to an older version and open an issue here.  I also don't think we've promised that the configuration interface to this package is covered under semver so I don't think we need a breaking change bump.  Finally, it also seems like _the right_ thing to do -- locking down which configuration values the underlying stream uses should make the client/agent behave in am ore stable fashion.

test/lib/utils.js
astorm: 👍 not too concerned about this check -- I'd be curious about the history of why it's here in the first place (but not SO curious that I'm going to track it down).

astorm (approved):
:+1: approving -- changes here seem reasonable and well thought it out.  Also, per 

> Before this change the keep-alive HTTP(S) agent would only be used if a second call to `client.config(...)` was made

It sounds like we've had scenarios where some users end up with agents that don't have keep-alive active, and other users end up with agents that DO have keep alive active.  This gives me confidence that APM Server can handle either scenario without issue, and that turning Keep-Alive permanently on won't be a problem.

index.js
trentm: If it helps later reviewing, the available Writable options are: https://nodejs.org/api/stream.html#stream_new_stream_writable_options

Thanks for bringing this up. The same potential "issue" exists for the options passed to the underlying HTTP Agent: https://nodejs.org/api/all.html#http_new_agent_options

I'll add a small note about these in the CHANGELOG.md -- which I need to update to sync with master anyway.

PR #140, opened 2021-02-25
chore: disallow package-lock.json file generation

trentm:
Also:
- anchors some .gitignore'd files to the root dir
- drops the "target" .gitignore because I don't know what that is meant to be from

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 /package-lock.json
 /.nyc_output
 /node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
-package-lock.json
-.nyc_output
-node_modules
-target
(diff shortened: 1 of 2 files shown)


astorm (approved):

watson:
The `target` folder came from this script: https://github.com/elastic/apm-nodejs-http-client/blob/e59d85db499a54f67772e61bf9ea07fda22ad117/.ci/scripts/convert_tap_to_junit.sh#L10

I don't think it's used anymore though

trentm:
@watson Thanks! It *is* being used in CI at least, so this was my bad. I'll add it back. (I've updated my `rg` usage to use `rg --hidden` so I search our ".ci/..." directories -- which is I missed this.)

trentm:
Re-added in commit 0beea9a

Issue #136, opened 2021-02-18
Blocking Behavior under Benchmarking Load


astorm:
We've received reports the client can block when both of the following conditions are true.

1. The instrumented application is under a heavy, artificial/benchmarking load 
2. The APM Server is in a non-response state

We've also been able [to reproduce this blocking behavior](https://github.com/astorm/elastic-experiments/tree/sdh-180) using a heavy [`autocannon`](https://www.npmjs.com/package/autocannon) run.

"Blocking" here means the entire application is blocked -- for example, the above express application will not response to HTTP requests when this blocking state is reached.  The application will typically block for 5 - 10 seconds before resuming normal operation.  

The intent of this issue is to diagnose this problem and collect any data from folks seeing this in their own benchmarking or production systems in order to identify the root cause, as well as identify configurations that may alleviate this issue.

astorm:
Capturing some context from a conversation elsewhere. 
> [Our current] assumption/educated-guess here is that it is: (a) buffers up all traces to send, (b) the client.write*() methods end up eventually being blocking when those buffers fill and (c) possibly (though we don't have evidence yet) there are bugs that result in map leaks if/when the APM server jam can be cleared.

trentm:
The issue did not appear (for me) when the demo app.js was handling only one request at a time (`autocannon -p 1`). That could mean it is still just a load issue, or it could mean there is a concurrency handling bug in the apm-server client.

astorm:
I spent this morning capturing some heap snapshots during these periods of an unresponsive APM server that leads to Node blocking.  Nothing conclusive to draw these these yet, but I did notice that over time the number of "objects allocated but _not_ released" appears to be growing in an unbounded way

<img width="1095" alt="early" src="https://user-images.githubusercontent.com/184372/108910330-8fc05900-75da-11eb-9f24-b36d3e38dd73.png" alt="Image showing 121,272 objects allocated with a 44.41MB heap size">

<img width="1095" alt="late" src="https://user-images.githubusercontent.com/184372/108910373-9d75de80-75da-11eb-8744-6a3a55a1f387.png" alt="Image showing 951,373 objects allocated with a 160MB heap size">


My intuition/working-theory is that something about the client's state during these period leads to a growing number of objects that can't be released _immediately_ and this eventually leads to a situation where the garbage collector starts trashing and blocking.  (if that sounds a little hand wavy, it is)

[...]

PR #142, opened 2021-03-08
9.6.0

trentm:
Doing a release to get #139 out separate from a (hopefully) coming change for #136. I probably won't update apm-agent-nodejs' package.json until there is a release with #136.

I bumped the minor version because while the change was just a bug fix, I feel it is a potential significant enough change in behaviour.


astorm (approved):
👍

trentm:
elastic-apm-http-client@9.6.0 tagged and published

PR #144, opened 2021-03-26
fix: blocking behaviour under load

trentm:
This make a few changes to fix issues with the APM agent under heavy
load and with a slow or non-responsive APM server.

1. A new `maxQueueSize` config option is added (default 1024 for now) to
   control how many events (transactions, spans, errors, metricsets)
   will be queued before being dropped if events are incoming faster
   than can be sent to APM server.

   Dev Note: This is a FIFO, so we will tend to drop new events and keep
   older ones, whose relevance fades. That isn't ideal. The current
   architecture of using node.js streams to buffer doesn't make it easy
   to drop older events. Dropping newer events under overload is deemed
   good enough for now.

2. JSON encoding of events (when uncorking) is done in limited size
   batches to control the amount of single chunk CPU eventloop blocking
   time. (See MAX_WRITE_BATCH_SIZE in Client._writev.) Internal stats
   are collected to watch for long(est) batch processing times.

   Dev Note: Eventually an investigation into putting JSON encoding into
   a separate worker thread might be interesting.

3. The handling of individual requests to the APM Server intake API has
   be rewritten to handle some error cases -- especially from a
   non-responsive APM server -- and to ensure that only one intake
   request is being performed at a time.

4. Support for backoff on intake API requests has been implemented per
   https://github.com/elastic/apm/blob/master/specs/agents/transport.md#transport-errors

Fixes: #136

* * *

This commit is a first carryover from https://github.com/trentm/apm-nodejs-http-client/tree/play
there is still work to be done before this is complete and reviewable.


trentm:
Some perf-ish results from testing [example-blocking-behavior.js](https://gist.github.com/trentm/187bd52c9c2178c9566ca3e608b9e175) with and without APM and talking to an APM server with different behaviours.

Typically (unless otherwise noted) we are hitting it with load via:

```
npx autocannon -d 90 -c 97 -p 10 localhost:3000/
```

which is: make requests as fast as possible for `90` seconds, using `97` concurrent connections, and `10` "pipelined requests". This is a higher level of concurrency in load than the [autocannon](https://github.com/mcollina/autocannon) 
default of `-c 10 -p 1`.

I also observed the memory usage ("REM") using htop while running each load test.


# tl;dr

Some observations from the data below:

1. In all load cases, the resident memory usage of the app stayed less than 150MB. The equivalent cases with an APM agent without these fixes would rise up to 1300MB.
2. This translated to much better latency figures for higher percentiles:
    ```
    │ Stat     │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev     │ Max     │
    ├──────────┼────────┼────────┼────────┼────────┼───────────┼───────────┼─────────┤
    │ no fix   │ 138 ms │ 155 ms │ 458 ms │ 665 ms │ 187.46 ms │ 237.45 ms │ 4876 ms │
    │ with fix │ 126 ms │ 145 ms │ 198 ms │ 208 ms │ 150.13 ms │  21.12 ms │  410 ms │
    ```
3. The "no APM server" case below demonstrates backoff working to not pepper the APM server with new requests when it is failing.
4. In all of the "using APM" cases, the internal APM agent stats show zero or a single APM event encoding event-loop bubble over 10ms. This is compared to without these fixes getting bubbles frequently under load, and up to over full seconds.


# baseline, no APM

htop observed: RES ~80MB, CPU ~104%

```
│ Stat    │ 2.5%  │ 50%   │ 97.5% │ 99%   │ Avg      │ Stdev   │ Max    │
├─────────┼───────┼───────┼───────┼───────┼──────────┼─────────┼────────┤
│ Latency │ 11 ms │ 17 ms │ 31 ms │ 36 ms │ 19.94 ms │ 7.39 ms │ 157 ms │
```


# with APM, local mockapmserver.js (accidentally with payloadLogFile)


```
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev    │ Max    │
├─────────┼────────┼────────┼────────┼────────┼───────────┼──────────┼────────┤
│ Latency │ 125 ms │ 138 ms │ 173 ms │ 180 ms │ 142.28 ms │ 18.25 ms │ 584 ms │

APM Agent stats: { apmclient:
   { numEvents: 611271,
     numEventsDropped: 597907,
     numEventsEnqueued: 13364,
     numEventsSent: 13364,
     slowWriteBatch:
      { numOver10Ms: 0,
        encodeTimeMs: 2.441588,
        fullTimeMs: 5.245072,
        numEvents: 32,
        numBytes: 25086 },
     backoffReconnectCount: 0 } }
```

# with APM, local mockapmserver.js that is accepting, reading, but not responding (accidentally with payloadLogFile)

htop: RES around 130-140MB

```
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev    │ Max    │
├─────────┼────────┼────────┼────────┼────────┼───────────┼──────────┼────────┤
│ Latency │ 126 ms │ 142 ms │ 181 ms │ 195 ms │ 145.81 ms │ 19.02 ms │ 584 ms │

APM Agent stats: { apmclient:
   { numEvents: 572840,
     numEventsDropped: 564596,
     numEventsEnqueued: 7316,
     numEventsSent: 5492,
     slowWriteBatch:
      { numOver10Ms: 0,
        encodeTimeMs: 2.647948,
        fullTimeMs: 5.601515,
        numEvents: 32,
        numBytes: 25085 },
     backoffReconnectCount: 3 } }
{"log.level":"error","@timestamp":"2021-03-29T21:38:23.224Z","log":{"logger":"elastic-apm-node"},"ecs":{"version":"1.6.0"},"message":"APM Server transport error: intake response timeout: APM server did not respond within 10s of gzip stream finish"}
```

# with APM agent, but no APM server (accidentally with payloadLogFile)

htop: RES around 130-140MB

```
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev    │ Max    │
├─────────┼────────┼────────┼────────┼────────┼───────────┼──────────┼────────┤
│ Latency │ 125 ms │ 137 ms │ 171 ms │ 178 ms │ 141.42 ms │ 17.01 ms │ 541 ms │
```


We see backoff delays, which is good: 4s, 9s, 16s, 25s

```
[2021-03-29T21:41:14.406Z] ERROR (elastic-apm-node): APM Server transport error (ECONNREFUSED): connect ECONNREFUSED 127.0.0.1:8200
[2021-03-29T21:41:14.761Z] ERROR (elastic-apm-node): APM Server transport error (ECONNREFUSED): connect ECONNREFUSED 127.0.0.1:8200
[2021-03-29T21:41:16.152Z] ERROR (elastic-apm-node): APM Server transport error (ECONNREFUSED): connect ECONNREFUSED 127.0.0.1:8200
[2021-03-29T21:41:20.260Z] ERROR (elastic-apm-node): APM Server transport error (ECONNREFUSED): connect ECONNREFUSED 127.0.0.1:8200
[2021-03-29T21:41:30.339Z] ERROR (elastic-apm-node): APM Server transport error (ECONNREFUSED): connect ECONNREFUSED 127.0.0.1:8200
[2021-03-29T21:41:47.043Z] ERROR (elastic-apm-node): APM Server transport error (ECONNREFUSED): connect ECONNREFUSED 127.0.0.1:8200
[2021-03-29T21:42:11.879Z] ERROR (elastic-apm-node): APM Server transport error (ECONNREFUSED): connect ECONNREFUSED 127.0.0.1:8200
```


# with APM agent, against APM server v7.12.0 in Google Cloud (us-west1)

htop: RES ~150MB

```
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev    │ Max    │
├─────────┼────────┼────────┼────────┼────────┼───────────┼──────────┼────────┤
│ Latency │ 126 ms │ 145 ms │ 198 ms │ 208 ms │ 150.13 ms │ 21.12 ms │ 410 ms │

APM Agent stats: { apmclient:
   { numEvents: 569401,
     numEventsDropped: 552964,
     numEventsEnqueued: 16085,
     numEventsSent: 15541,
     slowWriteBatch:
      { numOver10Ms: 0,
        encodeTimeMs: 5.624694,
        fullTimeMs: 6.186725,
        numEvents: 32,
        numBytes: 25084 },
     backoffReconnectCount: 0 } }
```

# comparing to client without these fixes

```
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev     │ Max     │
├─────────┼────────┼────────┼────────┼────────┼───────────┼───────────┼─────────┤
│ Latency │ 138 ms │ 155 ms │ 458 ms │ 665 ms │ 187.46 ms │ 237.45 ms │ 4876 ms │
```

htop: RES up to 1600MB



# with APM talking to cloud APM server, *with default autocannon load*

```
% npx autocannon -c 10 -d 90 -p 1 localhost:3000/
...
│ Stat    │ 2.5% │ 50%  │ 97.5% │ 99%  │ Avg     │ Stdev   │ Max   │
├─────────┼──────┼──────┼───────┼──────┼─────────┼─────────┼───────┤
│ Latency │ 1 ms │ 1 ms │ 5 ms  │ 5 ms │ 1.73 ms │ 1.04 ms │ 25 ms │


APM Agent stats: { apmclient:
   { numEvents: 392594,
     numEventsDropped: 54,
     numEventsEnqueued: 392540,
     numEventsSent: 387102,
     slowWriteBatch:
      { numOver10Ms: 1,
        encodeTimeMs: 2.460121,
        fullTimeMs: 12.261069,
        numEvents: 32,
        numBytes: 25088 },
     backoffReconnectCount: 0 } }
```

test/edge-cases.js
```diff
@@ -12,58 +12,6 @@ const assertMetadata = utils.assertMetadata
 const assertEvent = utils.assertEvent
 const validOpts = utils.validOpts
 
-test('Event: close - if ndjson stream ends', function (t) {
```
trentm, line 15: note to reviewer: These two tests were identical to the following two. I'm guessing they date back from before stream-chopper was in use, but I'm not sure.

index.js
```diff
@@ -528,80 +574,140 @@ Client.prototype._destroy = function (err, cb) {
...
     this._corkTimer = null
   }
-  clients[this._index] = null // remove global reference to ease garbage collection
+  clientsToAutoEnd[this._index] = null // remove global reference to ease garbage collection
   this._chopper.destroy()
   this._agent.destroy()
   cb(err)
 }
 
-function onStream (client, onerror) {
```
trentm, line 537: note to reviewer: A total re-write of `onStream` -- now called `getChoppedStreamHandler` and `makeIntakeRequest` --  is below (after the _getBackoffDelay() function). There is large block comment that should mostly explain it.

trentm (commented):
add some notes for reviewers.

index.js
```diff
@@ -610,36 +716,164 @@ function onStream (client, onerror) {
...
+      const backoffDelayMs = client._getBackoffDelay(!!err)
+      if (err) {
+        log.trace({ timeline, bytesWritten, backoffDelayMs, err },
+          'conclude intake request: error')
+        onerror(err)
+      } else {
+        log.trace({ timeline, bytesWritten, backoffDelayMs },
+          'conclude intake request: success')
+      }
+      if (backoffDelayMs > 0) {
```
astorm, line 728: If I understand this block correctly, the `if (backoffDelayMs > 0) {` is the code that delays the handling of the next request if there's a backoff delay. 

I see in one branch of the conditional we're calling `next()` directly in the current async frame/context (calling it synchronously), but that in the other branch we're scheduling `next` to run in a different async context via `setTimeout` (calling it asynchronously). Previous versions of the code always called next synchronously. 

Proforma question -- do we have any concerns about sometimes calling `next` synchronously and other times calling it asynchronously?

index.js
```diff
@@ -528,80 +574,140 @@ Client.prototype._destroy = function (err, cb) {
...
+    // the "error" and "close" events on a stream being called. When a part
+    // errors or all parts are completed, then we can conclude.
+    let concluded = false
+    const completedFromPart = {
+      gzipStream: false,
+      intakeReq: false,
+      intakeRes: false
+    }
+    let numToComplete = Object.keys(completedFromPart).length
+    const completePart = (part, err) => {
```
astorm, line 671: If I understand the intent of `completePart` -- it's a coordinating function that allows us to update the completed state of the gzip/chopper stream, the request, and the response.  I haven't completely traced out each event's code yet so this might be a naive question -- but is there any risk of a deadlock-ish condition here and is there anything we can/should do about that?

astorm (commented):
First review pass done -- nits and questions below. 

I'm going to take a second pass after I've given the client a try with a few toy applications  to make sure I understand the flow of things and to look for an unexpected/surprising changes in behavior we might need to address/document.

It's worth saying out loud that overall the new handler returned by `getChoppedStreamHandler` is more _locally_ complex than the previous `onStream` handler.  However, all we've really done here is surface the complexity that was hidden away under the `pump` library and given ourselves the necessary control to address the performance issues and edge cases.  It's complex code, but it's also as no more complex than it needs to be. 👍

test/lib/unref-client.js
```diff
@@ -1,8 +1,13 @@
 'use strict'
 
+// This is used in test/side-effects.js to ensure that a Client with a
+// (sometimes long-lived) request open to APM server does *not* keep a node
+// process alive.
+
 const Client = require('../../')
 
 const client = new Client({
+  // logger: require('pino')({ level: 'trace' }, process.stderr), // uncomment for debugging
```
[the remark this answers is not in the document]
trentm: Yes, ` // uncomment for debugging`.
Do you object to having something like this around? I'm fine removing it if so. Note that one would also need to `npm install --no-save pino` to turn this on.

index.js
```diff
@@ -1,5 +1,8 @@
 'use strict'
 
+const assert = require('assert')
```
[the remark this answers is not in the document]
trentm: Thanks for being open to trying it. I'm a strong "Yay", explicitly only using `assert` for cases where it is a bug in our code.  Looking at both cases where I've used it here:

```
assert(part in completedFromPart, `'${part}' is in completedFromPart`)
```

I feel this is an easy one. This is an internal function only called inside that enclosing `makeIntakeRequest`. It is just a bug in that code if it is called with a value other than the designated three part names. Attempting to make `completePart()` somehow cope with an invalid `part` value and stumble on with the intake request would make the code more complex and would tend to hide the programmer error in dev and testing.


```
    assert(client._encodedMetadata, 'client._encodedMetadata is set')
```

This is the one we discussed last time. Basically I make the same arguments: The code is made more complex attempting to cope with `client._encodedMetadata` not being set. While the code flow is less obvious compared to the `assert` above, there is no valid usage of the Client that allows `makeIntakeRequest` to be called before `client._encodedMetadata` is set. At that point, we, the developers don't understand the program state and can't make any guarantees about behaviour.

Another comment on the "code is made more complex": I was looking at completing the work for adding a filter on metadata (https://github.com/elastic/apm-agent-nodejs/issues/1916) on Monday. I decided to wait for this work to go in, because the current coping code here is a *4th* place where `client._encode(toEncode, Client.encoding.METADATA)` is called -- and where a new metadata filter should be applied.

In general, in my view, the important distinction is operational errors vs. programmer errors (as covered by https://www.joyent.com/node-js/production/design/errors).  These two cases cover what would be programmer errors. I realize the advice there:

> The best way to recover from programmer errors is to crash immediately.

is possibly controversial.  I understand APM Agent authors need to be very concerned about adding possible crashes in the user's app. However, I'd advocating for using asserts only to check internal invariants in the agent code to simplify and to makes subtle bugs obvious ones.

index.js
```diff
@@ -400,6 +432,7 @@ Client.prototype._writeFlush = function (cb) {
 
 Client.prototype._maybeCork = function () {
   if (!this._writableState.corked && this._conf.bufferWindowTime !== -1) {
+    // this._log.trace('cork (from _maybeCork)')
```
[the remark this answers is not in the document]
trentm: Thanks. Dropped in commit 84f7f57

index.js
```diff
@@ -528,80 +574,140 @@ Client.prototype._destroy = function (err, cb) {
...
+    const startTime = process.hrtime()
+    const timeline = []
+    let bytesWritten = 0
+    let intakeRes
+    let intakeResTimer = null
+    const intakeResTimeout = client._conf.intakeResTimeout
+    const intakeResTimeoutOnEnd = client._conf.intakeResTimeoutOnEnd
+
+    // `_active` is used to coordinate the callback to `client.flush(db)`.
+    client._active = true
```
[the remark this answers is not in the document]
trentm: Yup, the code is designed to only have one intake request active at a time. Otherwise there is a bug. The `completePart()` function is written to handle all the (documented in its top comment) ways the 3 streams can conclude, and only call StreamChopper's `next()` when those 3 streams are all cleanly finished or destroyed.

Note that this `client._active` handling was in the earlier `onStream` code. However, because the earlier code didn't wait on the completion of the HTTP response from APM server it *was* possible to have multiple intake requests going at the same time.

index.js
```diff
@@ -610,36 +716,164 @@ function onStream (client, onerror) {
...
+        // handling it here is preferable to an uncaughtException.
+        intakeResErr = wrapError(intakeResErr, 'intake response error event')
+        completePart('intakeRes', intakeResErr)
       })
-    }
+      intakeRes.on('data', (chunk) => {
+        chunks.push(chunk)
+      })
+      // intakeRes.on('close', () => { log.trace('intakeRes "close"') })
+      // intakeRes.on('aborted', () => { log.trace('intakeRes "aborted"') })
```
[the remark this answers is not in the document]
trentm: Again, if you are cool with it, I wanted to leave these in. They serve to:
- document the other possible events these streams might be emitting
- be available to turn on log.trace'ing if we find there is a bug down the line we want to investigate. For example, see the discussion further down about the "close" and "end" events on the gzipStream and the different behaviour I saw there for node v12 only.

index.js
trentm: That's fair. We could put the `next()` case in a `setImmediate(next)`. I'll do that.

Aside: I suppose we could have it always do `setTimeout(next, backoffDelayMs).unref()` even if backoffDelayMs was 0. Do you have a preference? They are both nominally the same, but libuv schedules them slightly differently (https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/).

index.js
trentm: 1. I guess the comment above "// Handle conclusion ..." isn't close enough to the function to obviously be a description for `completePart()`? If so, I'll move that comment block to be more obvious.
2. I don't believe there is a deadlock, no. There is always the `serverTimeout`, which starts a (default 30s) clock on idleness on the socket. As the comment further above states: `This is a catch-all fallback for an otherwised wedged connection.`.

trentm:
Testing notes:
- the "Test" check failure above is a spurious failure in the codecov.io reporting (https://github.com/elastic/apm-nodejs-http-client/issues/146)
- https://github.com/elastic/apm-agent-nodejs/pull/2024 demonstrates the APM agent test suite passing using these http client changes.

astorm:
jenkins run the tests please

test/lib/unref-client.js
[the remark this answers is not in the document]
astorm: Nope, no objection to keeping it around. Just wanted to make sure it was intentional. 👍

index.js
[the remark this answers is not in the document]
astorm: Sounds like a plan -- let's move forward with assertions for _programmer errors_ and not operational errors.  We can revisit the descision if/when we discover that the real world find ways to trigger our assertions.

index.js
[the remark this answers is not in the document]
astorm: 👍

index.js
astorm: No strong preference or enough practical experience to say which choice would be the better choice -- happy to defer this one.

index.js
astorm: The name and comment are fine -- _If I understand the intent_ is more about me than this code ;)

And 👍 on the no-deadlocks reasoning, resolving this comment.

index.js
```diff
@@ -466,32 +500,41 @@ Client.prototype._isUnsafeToWrite = function () {
   return this.destroyed
 }
 
+Client.prototype._shouldDropEvent = function () {
```
astorm, line 503: Do we want some sort of logging that indicates the client or agent is dropping events?

astorm (changes requested):
One last question before stamping this -- do you have any thoughts on whether we want to log a periodic message if/when the client ends up dropping spans?

I suspect there's going to be some span heavy applications/services that, with the previous version of the agent, didn't run into significant performance issues but who will end up hitting the span limit and dropping spans.   Do you think we need logging in the agent for this, or will the changelog(s) and documentation for `maxQueueSize` be enough?

index.js
trentm: From [earlier chat](https://elastic.slack.com/archives/C5M02D1GE/p1615912120010100):

One last thing I'm considering here is changing the agent to use process.emitWarning when we want to give warnings to the user rather than logger.warn() or logger.error(). That will:
1. allow us to not spam to user if it is a repeated situation (e.g. if we are dropping events because of load on the APM server)
2. make the text more readable for the user (it won't be a JSON log record)
3. allow the user to handle/silence particular warnings if they want

Sound reasonable if I defer that to a separate ticket?

trentm: https://github.com/elastic/apm-agent-nodejs/issues/2029 added to track this. Thanks for bringing it up.

astorm: > Sound reasonable if I defer that to a separate ticket?

👍

astorm (approved):
👍 thanks for talking those through.  GHA tests are passing, and the jenkins failures (per previous comments) appear to be related to the code coverage tooling.  Approving.

Issue #150, opened 2021-04-06
`client.flush(cb)` callback may not be called in v9.7.0


trentm:
The recent v9.7.0 release changed behaviour such that the `cb` callback in `client.flush(cb)` is not called in a specific case.

A minimal sample case with the APM agent (https://github.com/elastic/apm-agent-nodejs) is this:

```
require('elastic-apm-node').start({
  serviceName: 'use-agent-and-throw',
  centralConfig: false  // is necessary to trigger the issue
})

throw new Error('boom')
```

By default the elastic-apm-node agent is setting up an 'uncaughtException' handler that (eventually) effectively calls the following:

```
      client.sendError(error, function () {
        client.flush(function (err) {
          // ...
        })
      })
```

With previous versions of the http client, this would send the error to the APM server and call the callback to `client.flush()`. With v9.7.0, that client.flush() callback is not called. More details coming in subsequent comments.

trentm:
Here is a repro script just using the client (instead of via the apm-agent-nodejs). "call-me-back-maybe.js"

```js
// A script, used by test/side-effects.js, to test that the client.flush
// callback is called.
//
// We expect both `console.log`s to write their output.
//
// Two important things are required to reproduce the issue:
// 1. There cannot be other activities going on that involve active libuv
//    handles. For this Client that means:
//    - ensure no `_pollConfig` requests via `centralConfig: false`
//    - do not provide a `cloudMetadataFetcher`
// 2. There must be a listening APM server to which to send data.

const Client = require('../../') // elastic-apm-http-client

const serverUrl = process.argv[2]

const client = new Client({
  logger: require('pino')({ level: 'trace', ...require('@elastic/ecs-pino-format')() }, process.stderr), // uncomment for debugging
  serverUrl: serverUrl,
  serviceName: 'call-me-back-maybe',
  agentName: 'my-nodejs-agent',
  agentVersion: '1.2.3',
  userAgent: 'my-nodejs-agent/1.2.3',
  centralConfig: false // important for repro, see above
})

const e = { exception: { message: 'boom', type: 'Error' } }

client.sendError(e, function sendCb () {
  console.log('sendCb called')
  client.flush(function flushCb () {
    console.log('flushCb called')
  })
})
```

With that logger enabled for debugging, here is a run without a fix:

```
% node test/lib/call-me-back-maybe.js 2> >(ecslog -x process.pid,reqId)
[...]

PR #151, opened 2021-04-06
fix: ensure client.flush(cb) callback is called when no active handles

trentm:
Fixes: #150

Explanation of the issue and fix is on #150.

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,12 +46,7 @@ process.once('beforeExit', function () {
       // Clients remove themselves from the array when they end.
       return
     }
-    client._log.trace('ref client sockets and end client beforeExit')
-    // Calling _ref here, instead of relying on the _ref call in `_final`,
-    // is necessary because `client.end()` does *not* result in the Client's
-    // `_final()` being called when the process is exiting.
-    client._ref()
-    client.end()
+    client._gracefulExit()
   })
 })
 
@@ -78,6 +73,7 @@ function Client (opts) {
   this._configTimer = null
   this._encodedMetadata = null
   this._backoffReconnectCount = 0
+  this._intakeRequestGracefulExitFn = null // set in makeIntakeRequest
 
   // Internal runtime stats for developer debugging/tuning.
   this._numEvents = 0 // number of events given to the client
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -731,6 +748,19 @@ function getChoppedStreamHandler (client, onerror) {
       }
     }
 
+    // Provide a function on the client for it to signal this intake request
+    // to gracefully shutdown, i.e. finish up quickly.
+    client._intakeRequestGracefulExitFn = () => {
```
astorm, line 753: Thinking out loud about the concurrency here -- if I'm reading this correctly there's a period of time between one request finishing and another request starting where `client._intakeRequestGracefulExitFn` will "belong" to the previous request.  If the `process` object emits the `beforeExit` event during this period our handler will still call `_intakeRequestGracefulExitFn`.  However, this shouldn't have any ill effect as the `completePart` function has sufficient guard clauses and will bail out if the request is already finished.

test/lib/call-me-back-maybe.js
```diff
@@ -0,0 +1,34 @@
+// A script, used by test/side-effects.js, to test that the client.flush
```
astorm, line 1: [2012 called](https://www.google.com/search?client=firefox-b-1-d&q=call+me+maybe) ;)

astorm (approved):
Approving, and doing my "restating things in my own words" thing.

I see that we've modified our event handler for the `process` object's `beforeExit` handler and replaced the log-then-end code with a call to the new `_gracefulExit` method.

I see that `_gracefulExit` now checks for a `_intakeRequestGracefulExitFn` function on the client object, and if present, will call that method.  The `_gracefulExit` method also calls the client's `_ref` to ensure that each of the agent's sockets [is 'reffed](https://nodejs.org/api/net.html#net_socket_ref).

I see the `_intakeRequestGracefulExitFn` property is redefined whenever there's a new stream from the stream chopper to process (in `makeIntakeRequest`), and that when called it clears the current request's `intakeResTimer` timer and creates a new unrefed timer that will completed the `intakeRes` portion of `completePart` after 1 second (or whatever `intakeResTimeoutOnEnd` is set to).

index.js
trentm: Yes. Actually `intakeResTimer` should always be set null for that period. It might be more obvious to explicitly set this GracefulExitFn to null when concluding the request. Done in commit cab7b75

test/lib/call-me-back-maybe.js
trentm: Jepsen is canon. :)

PR #153, opened 2021-04-14
refactor: _fetchAndEncodeMetadata to not callback with args

trentm:
Nothing was using those callback args. Also clarify that an error from
getCloudMetadata is intentionally being ignored, and (lightly) log it.


astorm (approved):
👍

PR #154, opened 2021-04-15
feat: client.addMetadataFilter(fn)

trentm:
A facility to custom filter the "metadata" object sent to APM server at
the start of every intake request.

This refactors setting of `_encodedMetadata` to happen in only one
place (`_resetEncodedMetadata`), so that it can apply the metadata
filters in only on place.

Refs: https://github.com/elastic/apm-agent-nodejs/issues/1916

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,8 +263,11 @@ configuration options can be updated except:
 ### `client.addMetadataFilter(fn)`
 
 Add a filter function for the ["metadata" object](https://www.elastic.co/guide/en/apm/server/current/metadata-api.html)
-sent to APM server. Here is an example of a filter that removes the
-`metadata.process.argv` field:
+sent to APM server. This will be called once at client creation, and possibly
+again later if `client.config()` is called to reconfigure the client or
+`client.addMetadataFilter(fn)` is called to add additional filters.
+
+Here is an example of a filter that removes the `metadata.process.argv` field:
 
 ```js
 apm.addMetadataFilter(function dropArgv(md) {
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Arguments:
 - `options` - An object containing config options (see below). All options
   are optional, except those marked "(required)".
 
-Data sent to the APM Server as part of the metadata package:
+Data sent to the APM Server as part of the [metadata object](https://www.elastic.co/guide/en/apm/server/current/metadata-api.html):
 
 - `agentName` - (required) The APM agent name
 - `agentVersion` - (required) The APM agent version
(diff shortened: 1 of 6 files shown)


trentm:
The matching agent PR is https://github.com/elastic/apm-agent-nodejs/pull/2045

README.md
```diff
@@ -260,6 +260,37 @@ configuration options can be updated except:
...
+
+```
+[2021-04-14T22:28:35.419Z] ERROR (elastic-apm-node): APM Server transport error (400): Unexpected APM Server response
+APM Server accepted 0 events in the last request
+Error: validation error: 'metadata' required
+  Document: {"metadata":null}
+```
+
+See the [APM Agent `addMetadataFilter` documentation](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-add-metadata-filter)
+for further details.
```
astorm, line 291: saying it out loud -- [this](https://github.com/elastic/apm-agent-nodejs/pull/2045/files#diff-23427ac409c61f7de0225443c0c7c77aca90b732e73957bad73edb357c30addeR163) will need to land before this link works.

index.js
```diff
@@ -231,20 +237,43 @@ Client.prototype.config = function (opts) {
   // fixes bug where cached/memoized _encodedMetadata wouldn't be
   // updated when client was reconfigured
   if (this._encodedMetadata) {
-    this.updateEncodedMetadata()
+    this._resetEncodedMetadata()
```
astorm, line 240: Is it worth mentioning in the docs that filters will be applied _whenever_ the agent is reconfigured? (not just on startup)?

astorm (approved):
TL;DR; -- does what it says on the tin -- one question below about some possible docs additions, but approving.  Also, here's what I see i this PR

I see we've moved our logic for encoding the metadata from the `_fetchAndEncodeMetadata` method to a new `_resetEncodedMetadata` method.  The `_fetchAndEncodeMetadata` method is still responsible for initiating the _cloud_ metadata request, but all other work has been passed on to the `_resetEncodedMetadata` method. We've also added a new `_cloudMetadata` property in order to coordinate between `_fetchAndEncodeMetadata` and `_resetEncodedMetadata`.  The `_fetchAndEncodeMetadata` method sets this variable, and `_resetEncodedMetadata` is responsible for reading it. 

The new `_resetEncodedMetadata` method does a deep clone of the metadata stored at `this._conf.metadata` and, if present, the cloud metadata stored at `this._cloudMetadata`.   This deeply cloned object is passed through any callback/filter functions registered with the new `Filter/`object-filter-sequence` object.  The final filtered object is set as the `_encodedMetadata` object.

Speaking of filters, we've also added a public `addMetadataFilter` function that allows us to register a callback function with the `Filter/`object-filter-sequence` instance used by the client. 

Finally, the client invokes the `_resetEncodedMetadata` method in three places

1. After the initial fetch of the cloud metadata.
2. Whenever the client is reconfigured. 
3. When a filter is added IF `_encodedMetadata` has been set.

README.md
trentm: Yup.

index.js
trentm: Yup, that is worth it. Added in commit 9a12928

PR #159, opened 2021-06-02
perf: eliminate encodeObject stack and faster loop in _writeBatch

trentm:
A small perf refactor. _writeBatch is an inner function which can be
a hotspot under high event load.


astorm (approved):
Does the thing, approving. 👍

PR #160, opened 2021-06-09
9.8.1

trentm:
Get a release out with the small perf tweak, so we can use that in the agent.


astorm (approved):
👍

trentm:
elastic-apm-http-client@9.8.1 published

PR #162, opened 2021-08-31
feat: truncation changes

trentm:
- Truncate on number of chars, not bytes. The "maxLength" values in json schema (for the intake API) are [intended to be in JSON characters](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.3.1), i.e. unicode chars -- so this is a valid change. Truncating on chars (`str.length` check) is also *much* faster than the `unicode-byte-truncate` algorithm was.
- Remove truncateQueriesAt, add truncateLongFieldsAt which covers that
  and the other fields specified for [`long_field_max_length`](https://github.com/elastic/apm/blob/master/specs/agents/field-limits.md#long_field_max_length-configuration).
- Deprecate truncateErrorMessagesAt. If unspecified, the
  truncateLongFieldsAt value will be used.

BREAKING CHANGE. This "breaking" is for the http-client's API because `truncateQueriesAt` is being dropped, hence the bump to v10. However, it isn't a breaking change for the agent (at least not this part of it).

Refs: https://github.com/elastic/apm-agent-nodejs/pull/2193

### DIFF
--- a/lib/truncate.js
+++ b/lib/truncate.js
@@ -8,6 +8,37 @@ exports.span = truncSpan
 exports.error = truncError
 exports.metricset = truncMetricSet
 
+// Truncate the string `s` to a `max` maximum number of JavaScript characters.
+//
+// Note that JavaScript uses UCS-2 internally, so characters outside of the
+// BMP are represented as surrogate pairs. These count as *two* characters.
+// The result is that a string with surrogate pairs will appear to be truncated
+// shorter than expected:
+//    unitrunc('aaaa', 4) // => 'aaaa'
+//    unitrunc('😂😂😂😂', 4) // => '😂😂'
+//
+// This will avoid truncating in the middle of a surrogate pair by truncating
+// one character earlier. For example:
+//    unitrunc('foo😂bar', 4) // => 'foo'
+function unitrunc (s, max) {
+  if (s.length > max) {
+    if (max <= 0) {
+      return ''
+    }
+    // If the last character is a "high" surrogate (D800–DBFF) per
+    // https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Surrogates
+    // then we would truncate in the middle of a surrogate pair. Move back one
+    // char to have a clean(er) truncation.
+    const endChar = s.charCodeAt(max - 1)
+    if (endChar >= 0xd800 && endChar <= 0xdbff) {
+      return s.slice(0, max - 1)
+    } else {
+      return s.slice(0, max)
+    }
+  }
+  return s
+}
+
 function truncMetadata (metadata, opts) {
   return breadthFilter(metadata, {
     onArray,
@@ -97,10 +128,7 @@ function truncMetadata (metadata, opts) {
           break
       }
 
-      if (value.length > max) {
-        return value.slice(0, max)
-      }
-      return value
+      return unitrunc(value, max)
     }
   })
 }
@@ -132,10 +160,7 @@ function truncTransaction (trans, opts) {
           break
       }
 
-      if (value.length > max) {
-        return value.slice(0, max)
-      }
-      return value
+      return unitrunc(value, max)
     }
   })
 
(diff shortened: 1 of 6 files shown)


lib/truncate.js
```diff
@@ -272,7 +287,10 @@ function truncMetricSet (metricset, opts) {
         ? opts.truncateKeywordsAt
         : opts.truncateStringsAt
 
-      return truncate(value, max)
+      if (value.length > max) {
+        return value.slice(0, max)
```
astorm, line 291: Per discussion elsewhere, `slice` might have trouble with UTF-8 characters whose encoding requires more than three bytes. (🎉, etc.)

lib/truncate.js
```diff
@@ -362,7 +389,15 @@ function truncateCustomKeys (value, max, keywords) {
   }
   const result = value
   const keys = Object.keys(result)
-  const truncatedKeys = keys.map(k => keywords.includes(k) ? k : truncate(k, max))
+  const truncatedKeys = keys.map(k => {
```
astorm, line 392: 👍 for breaking up the one liner and adding the `if (k.length > max) {` check before truncating.

astorm (commented):
Looks like appropriate steps to take an a reasonable implementation.  Holding off on approval  until we've had a chance to try out the alternate truncation algorithms that could help with the `'foo🎉bar'.slice(0,4) === 'foo�'` issues.

lib/truncate.js
trentm: Yup, we are breaking surrogate pairs. Boo.
A truncation function (`unitrunc()`) was added in commit f3c179fe7a337a9616f54a33632c7d1c64a15dc8 that is still fast and doesn't break surrogate pairs.

Here is the test failing **without** unitrunc:

```
...
# do not break surrogate pairs in truncation
ok 491 context.db.statement was truncated
not ok 492 context.db.statement was truncated without breaking a surrogate pair
  ---
    operator: equal
    expected: 'foo'
    actual:   'foo�'
    at: Test.<anonymous> (/Users/trentm/el/apm-nodejs-http-client2/test/truncate.js:470:5)
    stack: |-
      Error: context.db.statement was truncated without breaking a surrogate pair
          at Test.assert [as _assert] (/Users/trentm/el/apm-nodejs-http-client2/node_modules/tape/lib/test.js:228:54)
          at Test.bound [as _assert] (/Users/trentm/el/apm-nodejs-http-client2/node_modules/tape/lib/test.js:80:32)
          at Test.equal (/Users/trentm/el/apm-nodejs-http-client2/node_modules/tape/lib/test.js:389:10)
          at Test.bound [as equal] (/Users/trentm/el/apm-nodejs-http-client2/node_modules/tape/lib/test.js:80:32)
          at Test.<anonymous> (/Users/trentm/el/apm-nodejs-http-client2/test/truncate.js:470:5)
          at Test.bound [as _cb] (/Users/trentm/el/apm-nodejs-http-client2/node_modules/tape/lib/test.js:80:32)
          at Test.run (/Users/trentm/el/apm-nodejs-http-client2/node_modules/tape/lib/test.js:96:10)
          at Test.bound [as run] (/Users/trentm/el/apm-nodejs-http-client2/node_modules/tape/lib/test.js:80:32)
          at Immediate.next [as _onImmediate] (/Users/trentm/el/apm-nodejs-http-client2/node_modules/tape/lib/results.js:83:19)
          at processImmediate (internal/timers.js:461:21)
  ...

1..492
# tests 492
# pass  491
# fail  1
```

astorm (approved):
Thanks for the `unitrunc` fix -- less corrupted unicode strings make the world a better place. Approving. 

(Also -- TIL _most_ of javascript is UCS-2/UTF-16ish internally and Node.js will occasionally sprinkle in some UTF-8 by default in its `Buffer` objects)

trentm:
elastic-apm-http-client@10.0.0 tagged and published

PR #163, opened 2021-09-02
tests: rename test files to *.test.js

trentm:
We did the same in the agent. It is nice to clarify/differentiate
from .js files that are *not* tests: like test utils and fixtures.


astorm (approved):
Looks like every file renamed is a test, that the `lib/*.js` files continue to have their old names, and that there's no other files lurking in the repo that `require('tape')`.  Approving.

PR #167, opened 2021-10-07
client improvements for Lambda

trentm:
- Update client.flush() to signal a current intake req to finish quickly.
  To support a lambda function invocation, an agent.flush(), at least for
  a sync lambda handler
  (https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html#nodejs-handler-sync),
  needs to socket.ref() so that the APM server communication is considered
  an active handle so that function execution continues.
  This also changes the APM server response timeout to the shorter value
  so that a slow APM server won't hang the function invocation for as
  long.
- The Lambda runtime sometimes uses the process 'beforeExit' event
  to decide when a function *invocation* is complete and *not* to
  indicate that the VM instance is done. The VM is frozen and the
  process may be re-used for subsequent invocations. Therefore, in
  Lambda we do not use 'beforeExit' to signal a graceful shutdown of the
  client.
- There was a race where an intake request could be started, then its
  `_intakeRequestGracefulExitFn()` called *before* that intake request
  had a socket to re-ref. This was triggered by the agent's lambda
  instrumentation that did `trans.end(); agent.flush()` in quick
  succession. If that `trans.end()` started an intake request, then the
  race was likely to be hit. The intake request's socket would then
  *not* get re-ref'd and the Lambda VM could get frozen before the
  APM data was flushed.

Fixes: #165 
Replaces: #166 

## Checklist

- [x] Changelog entry

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,7 +45,7 @@ const httpsRequest = https.request
 
 const containerInfo = getContainerInfo()
 
-const isLambdaExecutionEnviornment = !!process.env.AWS_LAMBDA_FUNCTION_NAME
+const isLambdaExecutionEnvironment = !!process.env.AWS_LAMBDA_FUNCTION_NAME
 
 // All sockets on the agent are unreffed when they are created. This means that
 // when the user process's event loop is done, and these are the only handles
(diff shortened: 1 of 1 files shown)


astorm (approved):
👍 approving.  I gave this client a try in a lambda function and it did not demonstrate any of the previous errors we saw.  Tested with both a callback based lambda and an async function/returned promise.

PR #170, opened 2021-11-01
10.3.0; expectExtraMetadata config & Client#setExtraMetadata

trentm:
These provide a mechanism for the Node.js APM Agent to pass in metadata
asynchronously and be sure that the client will not begin an intake
request until that metadata is provided.

Refs: https://github.com/elastic/apm-agent-nodejs/issues/2404

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,10 +92,9 @@ function Client (opts) {
   this._configTimer = null
   this._backoffReconnectCount = 0
   this._intakeRequestGracefulExitFn = null // set in makeIntakeRequest
-  // _encodedMetadata is pre-encoded JSON of metadata from `_conf.metadata` and
-  // `_cloudMetadata` (asynchronously fetched via `_conf.cloudMetadataFetcher`)
   this._encodedMetadata = null
   this._cloudMetadata = null
+  this._extraMetadata = null
   this._metadataFilters = new Filters()
 
   // Internal runtime stats for developer debugging/tuning.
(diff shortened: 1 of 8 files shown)


trentm:
I'll have a follow-up agent PR later today that uses this.

index.js
```diff
@@ -115,24 +114,35 @@ function Client (opts) {
...
-    // in case the client has been destroyed before this callback is called.
-    this._maybeUncork()
-
-    // the `cloud-metadata` event allows listeners to know when the
-    // agent has finished fetching and encoding its metadata for the
-    // first time
-    this.emit('cloud-metadata', this._encodedMetadata)
-  })
+  if (this._conf.cloudMetadataFetcher && this._conf.expectExtraMetadata) {
+    throw new Error('it is an error to create a Client with both cloudMetadataFetcher and expectExtraMetadata')
```
astorm, line 118: It looks like this will hard crash a service or application if it's misconfigured.  Are we OK with this?  My general understanding is we  (APM Agents generally) try to log a configuration but keep a customer's application or service.  Do we need a try/catch in the agent code now?

index.js
```diff
@@ -115,24 +114,35 @@ function Client (opts) {
...
+      // agent has finished fetching and encoding its metadata for the
+      // first time
+      this.emit('cloud-metadata', this._encodedMetadata)
+    })
+  } else if (this._conf.expectExtraMetadata) {
+    // Uncorking will happen in the expected `.setExtraMetadata()` call.
+    this._log.trace('corking (expectExtraMetadata)')
+    this.cork()
+  } else {
+    this._resetEncodedMetadata()
```
astorm, line 144: This call to `this._resetEncodedMetadata` wasn't present in the previous version of the object constructor and I don't completely follow what it's doing here now.  No need for a change, but I would like to understand what use case it handles or bug it fixes.

astorm (commented):
Generally good good -- it does the thing.  I've got a few questions before merging but once that are answered I should be good to approve.

index.js
trentm: It will intentionally throw if the Agent misuses this http client. I.e. if the Agent has a bug.  There isn't anything a customer's application code can do to impact that.

index.js
trentm: It handles the case where the Agent uses the client but passes neither `expectExtraMetadata`, nor `cloudMetadataFetcher`. This case is the empty "else"-block here: https://github.com/elastic/apm-agent-nodejs/pull/2407/files#diff-d5109920b34c2b2e141bd41232d8876432a17479c388e142d8a036bb7b36ea62R792-R800

Otherwise `this._encodedMetadata` is never set and the first intake request would fail.  For the other two cases `_resetEncodedMetadata` is called (a) when the cloudMetadataFetcher calls back or (b) when `client.setExtraMetadata()` is called.

index.js
astorm: I see -- so long as the agent never sends through this configuration the client won't throw, and we'll ensure that the agent doesn't do this. 👍

astorm (approved):

trentm:
elastic-apm-http-client@10.3.0 published
v10.3.0 tagged

PR #171, opened 2021-11-03
chore: Compression for intake requests should optimize for speed

trentm:


astorm (approved):
Seems like an OK thing to try.  FWIW, it looks like `Z_BEST_SPEED` gets its performance speed at the expense of compression efficiency. 

I gathered 723K of transactions into a file and tried compressing with the default level and comparing that to the `Z_BEST_SPEED` level.  The former compressed down to `38K`, while the later compressed down to `44K`.

PR #176, opened 2022-01-24
feat: Add APM Server version fetching and client.supportsKeepingUnsampledTransaction()

trentm:
This is to support the Node.js APM agent changing its behaviour based on
the APM Server version. Specifically whether to send or drop unsampled
transactions.

Refs: https://github.com/elastic/apm-agent-nodejs/issues/2455

### DIFF
--- a/index.js
+++ b/index.js
@@ -1052,16 +1052,25 @@ Client.prototype.supportsKeepingUnsampledTransaction = function () {
 Client.prototype._fetchApmServerVersion = function () {
   const self = this
 
-  const emitErrorAndSetUnknown = (errmsg) => {
+  const setVerUnknownAndNotify = (errmsg) => {
     self._apmServerVersion = null // means "unknown version"
     self.emit('request-error', new Error(errmsg))
+    if (isLambdaExecutionEnvironment) {
+      // In a Lambda environment, where the process can be frozen, it is not
+      // unusual for this request to hit an error. As long as APM Server version
+      // fetching is not critical to tracing of Lambda invocations, then it is
+      // preferable to not add an error message to the users log.
+      self._log.debug('verfetch: ' + errmsg)
+    } else {
+      self.emit('request-error', new Error(errmsg))
+    }
   }
   const headers = getHeaders(this._conf)
   // Explicitly do *not* pass in `this._agent` -- the keep-alive http.Agent
   // used for intake requests -- because the socket.ref() handling in
   // `Client#_ref()` conflicts with the socket.unref() below.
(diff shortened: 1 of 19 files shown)


trentm:
The first commit here is a first stab that has the APM Server fetching *retry* if it fails on earlier attempts. IIUC the java agent does *not* retry (see https://github.com/elastic/apm-agent-java/blob/master/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerHealthChecker.java). I think I am going to remove the retrying in the node.js agent, but I am not sure yet. If there are no retries, then if the APM Server happens to be down when the agent starts, then the agent will stay in a "degraded" state not knowing the APM server version.  If this "degraded" state just assumes the latest APM Server version then that is *perhaps* fine, but it could mean not getting data to the (older) APM server version if there is a compat issue.

index.js
```diff
@@ -1014,6 +1024,112 @@ function getChoppedStreamHandler (client, onerror) {
...
+ * Fetch the APM Server version and set `this._apmServerVersion`.
+ * https://www.elastic.co/guide/en/apm/server/current/server-info.html
+ *
+ * If fetching/parsing fails then the APM server version will be set to `null`
+ * to indicate "unknown version".
+ */
+Client.prototype._fetchApmServerVersion = function () {
+  const self = this
+
+  const setVerUnknownAndNotify = (errmsg) => {
```
astorm, line 1055: nit/question: With the `() => {}` style function the use of `self` shouldn't be necessary.  Any particular reason/mental-algorithm for making this an arrow function?

astorm (approved):
- I see the client now has a private property/member-variable named _apmServerVersion. 
- I see the value for _apmServerVersion will be read from the configuration provided when instantiating the client.
- If no configuration is provided when instantiating the client or the provided values is false-ish we'll call `Client._fetchApmServerVersion()` to fetch the value asyncronously from APM Server via an HTTP request
- A new method on the client, `supportsKeepingUnsampledTransaction`, uses the values to determine whether the APM Server versions supports keeping unsampled transactions.
- I see the functionality pulls in a new module dependency (`semver`,) for the client.  This seems reasonable, especially since we already use this module in the agent.
- The only potential hiccup I see is the period of time between instantiating a client and the request to APM server completing, but we appear to account for that by defaulting to the old, pre-APM Server 8 behavior.  

Approving.

index.js
trentm: This old dog is slowing attempting to master new tricks? ;)
I can drop the `self`ing.

trentm:
elastic-apm-http-client@10.4.0 published
v10.4.0 tagged

PR #178, opened 2022-02-24
feat: add support for Lambda data flushing

trentm:
This adds 'Client#lambdaStart()' and the 'lambdaEnd: true' option to
'Client#flush([opts,] [cb])' to support for flushing Lambda invocation
tracing data and signaling the Elastic Lambda extension per spec.
https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md#data-flushing

Refs: https://github.com/elastic/apm-agent-nodejs/issues/2485

# Checklist

- [x] waiting for [extension panic fixes](https://github.com/elastic/apm-aws-lambda/pull/132) before enabling the '?flushed=true' signaling by default
- [x] tests of the new APIs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # elastic-apm-http-client changelog
 
+## v10.5.0
+
+- Add support for coordinating data flushing in an AWS Lambda environment. The
+  following two API additions are used to ensure that (a) the Elastic Lambda
+  extension is signaled at invocation end [per spec](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md#data-flushing)
+  and (b) a new intake request is not started when a Lambda function invocation
+  is not active.
+
+  - `Client#lambdaStart()` should be used to indicate when a Lambda function
+    invocation begins.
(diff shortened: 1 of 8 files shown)


astorm (approved):
I don't really have a good mental model for how this client, particularly the flushing, works anymore so I can't really give a canonical :thumbsup: or :thumbsdown: on what overall effect this new code will have.  

However -- I've reviewed the new branching and from what I see the new Lambda flushing features appear to be gated correctly and leaves the non-lambda functionality intact.  In the few places where the mainline functionality has been altered (ex. addition of the flush marker to `_writeFlush`) the changes are minor enough that it shouldn't effect the overall behavior of the client when operating in non-lambda mode.

Also, for what it's worth, this is another example where a separate/different HTTP client (that implements the same defacto-public interface) for Lambda would make this sort of functionality less-risky/less-difficult to implement.  It may make sense in the future to take advantage of the Agent's architecture around it's transport object.  Not recommending this change be made now, just dinging that bell again. 

Approving.

trentm:
elastic-apm-http-client@11.0.0 is published

PR #179, opened 2022-03-07
fix: occasional "socket hang up" errors in APM server intake requests

trentm:
Use the 'agentkeepalive' alternative to Node.js core `http.Agent`
for keep-alive handling to get the `freeSocketTimeout` option that
can timeout kept-alive (aka "free") sockets before getting in the
range of the downstream APM Server (or Lambda extension) HTTP
Keep-Alive timeout. This avoids occasional ECONNRESET errors.

Fixes: #2594

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # elastic-apm-http-client changelog
 
+## Unreleased
+
+- Add the `freeSocketTimeout` option, with a default of 4000 (ms), and switch
+  from Node.js's core `http.Agent` to the [agentkeepalive package](https://github.com/node-modules/agentkeepalive)
+  to fix ECONNRESET issues with HTTP Keep-Alive usage talking to APM Server
+  (https://github.com/elastic/apm-agent-nodejs/issues/2594).
+
 ## v10.4.0
 
 - Add APM Server version checking to the client. On creation the client will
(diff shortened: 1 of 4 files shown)


test/edge-cases.test.js
```diff
@@ -347,7 +347,7 @@ test('socket timeout - server response too slow', function (t) {
       t.ok(err, 'got a request-error from the client')
       const end = Date.now()
       const delta = end - start
-      t.ok(delta > 1000 && delta < 2000, 'timeout should occur between 1-2 seconds')
+      t.ok(delta > 1000 && delta < 2000, `timeout should occur between 1-2 seconds: delta=${delta}ms`)
```
astorm, line 350: +1 for some better test logging.

astorm (approved):
Swapping in a new http.Agent seems like a pretty big change, but if we are going to do it `agentkeepalive` looks like a reasonable choice.  It's a  long standing project with a maintainer that's still semi-engaged.  And if it fixes some lambda problems then I guess we're in "trust our tests and <s>yolo</s> be agile" territory.

The specific code changes themselves appear to do what they intend to.  

I see that we're ensuring `keepAliveMsecs`, `maxSockets`, `maxFreeSockets`, and `freeSocketTimeout` are set on our configuration object.

I see we're conditionally choosing which keep alive constructor-function/class to use (the HTTP or HTTPS version) based on the protocol portion of the server URL.

I see we're then instantiating an object with that constructor function/class and including two new configuration fields -- `freeSocketTimeout` and `serverTimeout`.

Approving.

PR #180, opened 2022-03-09
fix: do not back-off on intake req errors in Lambda env

trentm:
Doing the back-off in a Lambda env can result in the APM agent instrumentation *breaking* the Lambda function response: https://github.com/elastic/apm-agent-nodejs/issues/2598
See the inline comment and the motivation at: https://github.com/elastic/apm/pull/613

## commit message

```
fix: do not back-off on intake req errors in Lambda env

Refs: https://github.com/elastic/apm-agent-nodejs/issues/2598
Refs: https://github.com/elastic/apm/pull/613
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # elastic-apm-http-client changelog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


astorm (approved):
👍 approved

PR #183, opened 2022-09-30
feat: container-info import and ECS Container Information

astorm:
Closes: https://github.com/elastic/apm-agent-nodejs/issues/2914

This PR imports the `container-info` module into this repo as a private library along with its tests.

It also modifies this module to include the `_` to `-` renaming that we'd previously done indirectly. 

It also modifies this module to use the ECS Fargate regular expression when the ECS metadata file is not present. 

It also modifies the k8s.test.js suite to use the new library structure.

Finally, it refactors the client to use the new local `container-info` library.


lib/container-info.js
```diff
@@ -1,20 +1,103 @@
...
- *
- * https://github.com/elastic/apm/blob/main/specs/agents/metadata.md#containerkubernetes-metadata
- */
-module.exports = (getInfo = getContainerInfo) => {
-  const info = getInfo.sync()
-  if (!info) { return info }
-
-  if (info.podId) {
-    info.podId = info.podId.replace(/_/g, '-')
+'use strict'
```
astorm, line 1: NOTE: Original is here if diffs are required: https://github.com/Qard/container-info/blob/v1.1.0/index.js

test/container-info.test.js
```diff
@@ -0,0 +1,423 @@
+'use strict'
```
astorm, line 1: NOTE: original here: https://github.com/Qard/container-info/blob/v1.1.0/test/test.js

test/fixtures/cgroup
```diff
@@ -0,0 +1,14 @@
+14:pids:/kubepods/kubepods/besteffort/pod0e886e9a-3879-45f9-b44d-86ef9df03224/244a65edefdffe31685c42317c9054e71dc1193048cf9459e2a4dd35cbc1dba4
```
astorm, line 1: NOTE: original here: https://github.com/Qard/container-info/blob/v1.1.0/test/fixtures/cgroup

test/fixtures/cgroup_result.js
```diff
@@ -0,0 +1,120 @@
+'use strict'
```
astorm, line 1: NOTE: original here: https://github.com/Qard/container-info/blob/v1.1.0/test/fixtures/result.js

test/fixtures/ecs-container-metadata.json
```diff
@@ -0,0 +1,3 @@
+{
```
astorm, line 1: Note: original here: https://github.com/Qard/container-info/blob/v1.1.0/test/fixtures/ecs-container-metadata.json

astorm:
Should merge with https://github.com/elastic/apm-nodejs-http-client/pull/184 when it lands before proceeding in order to ensure there are no regressions.

lib/container-info.js
```diff
@@ -1,20 +1,103 @@
...
+function ecsMetadataSync (ecsMetadataFile) {
+  try {
+    return ecsMetadataFile && JSON.parse(fs.readFileSync(ecsMetadataFile))
+  } catch (err) {}
+}
+
+module.exports = containerInfo
+containerInfo.sync = containerInfoSync
+containerInfo.parse = parse
+containerInfo.resetEcsMetadata = resetEcsMetadata
```
trentm, line 103:
```suggestion
containerInfo.resetEcsMetadata = resetEcsMetadata // Exported for testing-only.
```

lib/container-info.js
```diff
@@ -1,20 +1,103 @@
...
+  if (podId) data.podId = podId.replace(/_/g, '-')
+
+  const taskId = (path.match(taskReg) || [])[1]
+  if (taskId) data.taskId = taskId
+
+  // if we reach the end and there's still no conatinerId match
+  // and there's not an ECS metadata file, try the ECS regular
+  // expression in order to get a container id in fargate
+  if (!containerId && !ecsMetadata) {
+    const basename = path.split('/').pop()
```
trentm, line 43: Perhaps `.pop()` the `parts` array above into `basename` above so we don't need to `path.split('/')` twice.

trentm (commented):
Looks good. Holding approval until we've discussed the side issue.

lib/container-info.js
astorm: Sure, small enough change.  Done in 9f7521f5a98c3058f39f42924061a51aa15ea817. 

After a few years of working with you I realize your approach is very "let's gut what's there while we can" while I'm very "let's not touch that old fuse box because who knows what will catch on fire" :)

trentm (approved):

PR #185, opened 2022-10-24
fix: put [5s, 1d] range guard on the delay before re-fetching central config

trentm:
Refs: https://github.com/elastic/apm-agent-nodejs/issues/2941

The [spec change](https://github.com/elastic/apm/pull/667/files) requires a minimum 5s delay for refetching central config. The Node.js APM agent didn't have the "spin loop fetching central config from `Cache-Control: max-age=0`" issue. However, theoretically a *large* max-age value could overflow the max `setTimeout` delay (https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value) which could result in a similar spin loop. I picked a maximum value of 1 day -- a large enough value that I cannot imagine it getting in a way of a legitimate max-age value from the APM server.

It might be worth adding a max guard to the spec.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,6 +25,7 @@ const StreamChopper = require('stream-chopper')
 const ndjson = require('./lib/ndjson')
 const { NoopLogger } = require('./lib/logging')
 const truncate = require('./lib/truncate')
+const { getCentralConfigIntervalS } = require('./lib/central-config')
 
 module.exports = Client
 
@@ -432,16 +433,7 @@ Client.prototype._pollConfig = function () {
 Client.prototype._scheduleNextConfigPoll = function (seconds) {
   if (this._configTimer !== null) return
 
-  // Re-fetch central config after the given number of `seconds`.
-  // Default to 5 minutes, minimum 5s, max 1d.
-  //
-  // The maximum of 1d ensures we don't get surprised by an overflow value to
-  // `setTimeout` per https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value
-  const DELAY_DEFAULT_S = 300 // 5 min
-  const DELAY_MIN_S = 5
-  const DELAY_MAX_S = 86400 // 1d
-  const delayS = Math.min(Math.max(seconds || DELAY_DEFAULT_S, DELAY_MIN_S), DELAY_MAX_S)
-
+  const delayS = getCentralConfigIntervalS(seconds)
   this._configTimer = setTimeout(() => {
     this._configTimer = null
     this._pollConfig()
--- a/index.js
+++ b/index.js
@@ -432,12 +432,20 @@ Client.prototype._pollConfig = function () {
 Client.prototype._scheduleNextConfigPoll = function (seconds) {
   if (this._configTimer !== null) return
 
-  seconds = seconds || 300
+  // Re-fetch central config after the given number of `seconds`.
+  // Default to 5 minutes, minimum 5s, max 1d.
+  //
+  // The maximum of 1d ensures we don't get surprised by an overflow value to
+  // `setTimeout` per https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1464,9 +1472,10 @@ function normalizeGlobalLabels (labels) {
   return result
 }
 
+// https://httpwg.org/specs/rfc9111.html#cache-response-directive.max-age
 function getMaxAge (res) {
   const header = res.headers['cache-control']
-  const match = header && header.match(/max-age=(\d+)/)
+  const match = header && header.match(/max-age=(\d+)/i)
```
trentm, line 1478: Review note: https://httpwg.org/specs/rfc9111.html#field.cache-control says (emphasis mine):

> Cache directives are identified by a token, **to be compared case-insensitively** ...

astorm (approved):
:+1: Does the thing and preserves most of the old behavior for the edge case arguments (`0`, other false-y values result in `300` -- `true` now comes through at `5` instead of `1`, which is inline with the change).  Strings still result in `NaN` (per the previous algorithm), which is a little weird but it's not a public API so that doesn't seem like an edge case worth fixing.  

```
    +--------------------------------+
    | Sec. Input  | Calculated Delay |
    +--------------------------------+
    | -4          | 5                | 
    | -3          | 5                | 
    | -2          | 5                | 
    | -1          | 5                | 
    | 0           | 300              | 
    | 1           | 5                | 
    | 2           | 5                | 
    | 3           | 5                | 
    | 4           | 5                | 
    | 5           | 5                | 
    | 6           | 6                | 
    | 7           | 7                | 
    | 8           | 8                | 
    | 9           | 9                | 
    | 10          | 10               | 
    | 86398       | 86398            | 
    | 86399       | 86399            | 
    | 86400       | 86400            | 
    | 86401       | 86400            | 
    | 86402       | 86400            | 
    | 86403       | 86400            | 
    | 86404       | 86400            | 
    | null        | 300              | 
    | undefined   | 300              | 
    | FALSE       | 300              | 
    | TRUE        | 5                | 
    | “a string”  | NaN              | 
    +--------------------------------+
```

trentm:
> Strings still result in `NaN` (per the previous algorithm), which is a little weird

Agreed. Thanks. I've improved it to return the default value for any non-positive number. From a test run:

```
# getCentralConfigIntervalS
ok 1 getCentralConfigIntervalS(-4) -> 300
ok 2 getCentralConfigIntervalS(-1) -> 300
ok 3 getCentralConfigIntervalS(0) -> 300
ok 4 getCentralConfigIntervalS(1) -> 5
ok 5 getCentralConfigIntervalS(2) -> 5
ok 6 getCentralConfigIntervalS(3) -> 5
ok 7 getCentralConfigIntervalS(4) -> 5
ok 8 getCentralConfigIntervalS(5) -> 5
ok 9 getCentralConfigIntervalS(6) -> 6
ok 10 getCentralConfigIntervalS(7) -> 7
ok 11 getCentralConfigIntervalS(8) -> 8
ok 12 getCentralConfigIntervalS(9) -> 9
ok 13 getCentralConfigIntervalS(10) -> 10
ok 14 getCentralConfigIntervalS(86398) -> 86398
ok 15 getCentralConfigIntervalS(86399) -> 86399
ok 16 getCentralConfigIntervalS(86400) -> 86400
ok 17 getCentralConfigIntervalS(86401) -> 86400
ok 18 getCentralConfigIntervalS(86402) -> 86400
ok 19 getCentralConfigIntervalS(86403) -> 86400
ok 20 getCentralConfigIntervalS(86404) -> 86400
ok 21 getCentralConfigIntervalS(null) -> 300
ok 22 getCentralConfigIntervalS(undefined) -> 300
ok 23 getCentralConfigIntervalS(false) -> 300
ok 24 getCentralConfigIntervalS(true) -> 300
ok 25 getCentralConfigIntervalS(a string) -> 300
ok 26 getCentralConfigIntervalS([object Object]) -> 300
ok 27 getCentralConfigIntervalS() -> 300
```

PR #187, opened 2022-11-03
Add type information for typescript

nicolaichuk:
@trentm Please review and merge to upstream.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,842 @@
+/// <reference types="node" />
+
+// ///////////////////////////////////////////////////////////////////////////
+// Logger
+
+import { EventEmitter } from "events"
+
+/** @see https://github.com/pinojs/pino/blob/master/pino.d.ts */
+
+export declare interface LogFn {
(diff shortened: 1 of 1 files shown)


trentm:
@nicolaichuk Thanks for the PR, and my apologies for the long time in replying. I don't want to take a large .d.ts file for this package because:
- It would be a maintenance burden, for example to keep the `Transaction`, `Span`, `Error`, `Metricset` types in sync with the schema from https://github.com/elastic/apm-server/tree/main/docs/spec/v2.
- My expectation is that the only direct user of this repo is the APM agent at https://github.com/elastic/apm-agent-nodejs. I may be mistaken, but I expect there are very few users that benefit from using the apm HTTP client directly.  (Recently in https://github.com/elastic/apm-agent-nodejs/pull/3507 this http-client package was moved into the elastic/apm-agent-nodejs repo and the plan is to archive this repo.)

We would possibly consider exporting times for the http-client in that repo, but the same maintenance issue with such a well-defined, large .d.ts file would still apply.

PR #188, opened 2022-11-15
fix: Prevent a possible tight loop in central config fetching.

trentm:
If a central config request does not receive a Cache-Control response
header with a `max-age` field, then it incorrectly results in `NaN` for
the interval for the next central config fetch.

Refs: https://github.com/elastic/apm-agent-nodejs/issues/3029

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # elastic-apm-http-client changelog
 
+## v11.0.3
+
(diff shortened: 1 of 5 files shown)


astorm (approved):
All good here for the short term. 

Long term I'm wondering if we should start adding formal unit tests for every method we're going to change in the client with all the standard types and psuedo-types (NaN, undefined, null, etc.)  This one slipped by us both.

PR #189, opened 2022-11-23
feat: add an 'agentActivationMethod' config var

trentm:
This adds 'metadata.service.agent.activation_method' if the optional
config var is provided. Bump to v11.2.0.

Refs: https://github.com/elastic/apm-agent-nodejs/issues/3039


trentm:
elastic-apm-http-client@11.2.0 published
v11.2.0 tagged

PR #190, opened 2022-11-28
fix: change default APM server_url to 'http://127.0.0.1:8200' to avoid IPv6 ambiguity

trentm:
The APM agent won't *require* this change, because in the referenced
change it will be adding the same default on its end.

Refs: https://github.com/elastic/apm-agent-nodejs/pull/3049

# Checklist

- [x] Wait for the APM spec PR to get some feedback: https://github.com/elastic/apm/pull/727

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # elastic-apm-http-client changelog
 
+## v11.0.4
+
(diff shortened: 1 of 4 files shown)


trentm:
elastic-apm-http-client@11.0.4 published with this

PR #192, opened 2022-12-15
feat: add 'extraMetadata' config option

trentm:
This is a way to provide additional metadata that is merged into the
metadata object created from the other individual config options. This
is similar to `expectExtraMetadata: true` && `setExtraMetadata(...)`,
but is a way to provide the metadata *synchronously*.

Refs: https://github.com/elastic/apm-agent-nodejs/pull/3071

* * *

This'll be used for the Azure Functions work. Like Lambda instrumentation, some extra metadata is provided separate from the "cloud metadata fetcher" mechanism. However, unlike Lambda we can do this synchronously without needing to way for the first Function invocation because all data we need is in the environment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # elastic-apm-http-client changelog
 
+## v11.1.0
+
+- Add an `extraMetadata` config option, which is an object to merge into the
+  built metadata object. This is an alternative to the existing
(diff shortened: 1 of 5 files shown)


trentm:
`elastic-apm-http-client@11.1.0` published, v11.1.0 tagged

PR #198, opened 2023-04-03
v11.3.1; default to sending `activation_method` field if APM server version is not (yet) known

trentm:
Optimistically assume the APM server version is not v8.7.0. This allows
typically avoiding the v8.7.0 bug (which isn't that severe) with still
sending the `activation_method` for intake requests before the APM
server version fetch completes -- e.g. for fast first serverless
invocations.

Refs: https://github.com/elastic/apm/issues/781

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # elastic-apm-http-client changelog
 
+## v11.3.1
+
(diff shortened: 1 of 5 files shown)


trentm:
elastic-apm-http-client@11.3.1 published to npm

PR #200, opened 2023-06-20
feat!: metadata.system.{configured_hostname,detected_hostname}, prefer FQDN

trentm:
This is a breaking change, and bumps to version 12.0.0.

- Also belatedly add testing for node v20.

Refs: https://github.com/elastic/apm-agent-nodejs/issues/3310


index.js
```diff
@@ -1245,6 +1250,13 @@ Client.prototype.supportsActivationMethodField = function () {
     return semver.gte(this._apmServerVersion, '8.7.1')
   }
 }
+Client.prototype.supportsConfiguredAndDetectedHostname = function () {
+  if (!this._apmServerVersion) {
+    return true // Optimistically assume APM server is >=7.4.
```
trentm, line 1255: Reviewer note: I'm considering changing this, but I'm not sure. It would be *nice* to just be able to assume we are on newer APM server... but I suppose for someone using APM server <7.4 *and* when fetching the APM server version fails (for whatever reason), we will no longer be sending `metadata.system.hostname`. I don't think that'll *break* anything, but it would be a change in the data. Thoughts?

test/edge-cases.test.js
```diff
@@ -446,6 +446,7 @@ test('client.destroy() - on client with request in progress', function (t) {
   server.listen(function () {
     client = new Client(validOpts({
       serverUrl: 'http://localhost:' + server.address().port
+      // TODO: the _fetchApmServerVersion() here *is* hanging.
```
trentm, line 449: It would be good to deal with this in a separate change. This test case hangs for 30s for the timeout in `_fetchApmServerVersion()`.

trentm:
@david-luna Please review when you have a chance. This is still in draft while the spec PR (https://github.com/elastic/apm/pull/805) is being discussed. Then when this is in, I'd do a release and update the agent PR (https://github.com/elastic/apm-agent-nodejs/pull/3437).

index.js
david-luna: Not a GO developer but looking at [the changes in apm-server](https://github.com/elastic/apm-server/pull/2540/files#diff-db6b9d7c7ae6d368324efedf21092eca303205602ee5b259e737837b276f0cd1L27) if we sent `metadata.system.hostname` the server is going to drop it. So by returning `true` we are implicitly encouraging to update apm-server which IMO is good.

david-luna (approved):

index.js
trentm: Meh, I think it is fine as is.

trentm:
"v12.0.0" tagged and elastic-apm-http-client@12.0.0 published to npm

